PR #16, opened 2023-10-02
Add Discord RPC

svalencia014:
This PR adds discord rich presence and some slight gitignore changes to prevent the Macos binary from being checked in.


mmp:
Looks great overall! I will do a closer review and merge it tomorrow.

main.go
```diff
@@ -208,6 +210,26 @@ func main() {
 			uiShowConnectDialog(false)
 		}
 
+		//Initialize discord RPC
+		now = time.Now()
```
mmp, line 214: Two comments about `now`:
1. it's not super descriptive, since it's really more like the start time, so maybe rename to something like `simStartTime`?
2. If I'm following the code correctly, it's only set once, while it probably should be reset whenever the user either starts a new sim or quits from the current one to give a more accurate elapsed time. If you reset it immediately after the below line `case nw := <-newWorldChan:` (around line 241, thanks github for not letting me add a comment there), then that should take care of that.

world.go
```diff
@@ -562,6 +563,16 @@ func (w *World) GetWindowTitle() string {
 		return "(disconnected)"
 	} else {
 		deparr := fmt.Sprintf(" [ %d departures %d arrivals ]", w.TotalDepartures, w.TotalArrivals)
+		//Update discord RPC
+		client.SetActivity(client.Activity{
```
mmp, line 567: These `client.` calls are a little ambiguous in the vice context since it has a bunch of code elsewhere that uses `client` based variables (mostly around multi-user support.)

It might be better to import the `rich-go/client` stuff like:
```
  discord_client "github.com/hugolgst/rich-go/client"
```
which is go's way of letting you give a package new name. Then this call (and the calls in main.go) would be `discord_client.SetActivity(...` which would make it more clear what they were doing. (Noted that you have a comment above here, which helps, though.)

mmp, line 567: Another comment is that this will send an update to discord every frame, so 30-60 times per second. It's probably not a ton of bandwidth but on the other hand, it's not sending any information.

I think it might work to only send this update when a new sim starts or finishes (right after the `case nw := <-newWorldChan:` line in main.go), which would reduce it to the minimum number of updates needed to send to discord..

svalencia014: I have placed the update there since it will provide accurate numbers for current departures and arrivals. Would there be another place that would provide timely departure and arrival rates that is not as fast as every frame?

mmp: Good point about arrivals/departures. One option would be to do the update in the code where new departures/arrivals are created, but that's kind of obscure; it seems more clear to have the discord stuff happening somewhere easier to find; this feels like the right spot for it in general.

How about this? In main(), add a `lastDiscordUpdate := time.Now()` variable before the main loop (around line 239 in your updated version). Then inside that `for {` loop, add something like:

```
if time.Since(lastDiscordUpdate) > 5 * time.Second {
  // update discord. if world == nil, then it's disconnected...
  lastDIscordUpdate = time.Now()
}
```

then it will send regular updates but not too often.

mmp (changes requested):
Another thing that would probably be good would be to clear the discord status when vice exits. That would best be done inside the `if platform.ShouldStop() && len(ui.activeModalDialogs) == 0 {` test at the bottom of main.go.

mmp:
A last comment is that the user should probably have some control over this (and the option to opt-in/opt-out. That can wait for a followup, though.

svalencia014:
> Another thing that would probably be good would be to clear the discord status when vice exits. That would best be done inside the `if platform.ShouldStop() && len(ui.activeModalDialogs) == 0 {` test at the bottom of main.go.

I'll add that in, although the library I implemented appears to automatically terminate the connection when the program exits

world.go
```diff
@@ -562,6 +563,16 @@ func (w *World) GetWindowTitle() string {
 		return "(disconnected)"
 	} else {
 		deparr := fmt.Sprintf(" [ %d departures %d arrivals ]", w.TotalDepartures, w.TotalArrivals)
+		//Update discord RPC
```
mmp, line 566: So I think this part still needs to move to main.go, under `if time.Since(lastDiscordUpdate) > 5 * time.Second`, for the case where `world != nil` and

mmp:
Looks good! There's some UI to add to let users enable/disable this and also to opt in/out when they first launch vice. If you're interested in doing that let me know and I can offer some thoughts / give some pointers about how to add it.

PR #18, opened 2023-11-17
Add video maps

jessie846:
Add the compressed video map for files for every ARTCC, in an attempt to save time and confusion for other developers when they can't figure out how to convert the video maps, or do not have CRC.


mmp:
Good idea and sorry for sitting on this! Merged in 3b27770aba4c8eec576.

PR #20, opened 2024-01-13
Removed EWR scratchpad

checkandmate1:
Newark IRL doesn't use scratchpads. Removed it from the file for ILS 22L NOA + NFI & ILS 4R ARD + NFI


mmp:
Thanks!

PR #22, opened 2024-01-17
Live wind 

checkandmate1:
Added live wind. The package (checkandmate/AirportWeatherData) will pull the airports METAR from aviationweather.gov and provide the wind direction, speed and gust. The live wind checkbox will disable if:

- KAAC is selected
- No internet connection is found

### DIFF
--- a/resources/scenarios/ind.json
+++ b/resources/scenarios/ind.json
@@ -1,0 +1,1643 @@
+{
+  "tracon": "IND (Indianapolis)",
+  "airports": {
+    "KIND": {
+      "name": "Indianapolis",
+      "approaches": {
+        "I5L": {
+          "full_name": "ILS Runway 5L",
+          "runway": "05L",
+          "tower_controller": "IND_E_TWR",
+          "type": "ILS",
+          "waypoints": [
+            "KELLY/a5000 WAAYN/a5000/s210 FTBLL/a4000 STDUM/a3000 OINKK/a2100 OSCRR/a1220 _IND_05L/a747"
+          ]
+        },
+        "R5L": {
+          "full_name": "RNAV Y Runway 5L",
+          "runway": "05L",
+          "tower_controller": "IND_E_TWR",
+          "type": "RNAV",
+          "waypoints": [
+            "KELLY/a5000 WAAYN/a5000/s210 FTBLL/a4000 STDUM/a3000 OINKK/a2100 _IND_05L/a747"
+          ]
+        },
+        "I5R": {
+          "full_name": "ILS Runway 5R",
+          "runway": "05R",
+          "tower_controller": "IND_W_TWR",
+          "type": "ILS",
+          "waypoints": [
+            "KELLY/a5000 IIUUU/a5000/s210 BIGRD/a4000 BGTEN/a3000 HUUPS/a2100 GOLSE/a1240 _IND_05R/a791"
+          ]
+        },
+        "R5R": {
+          "full_name": "RNAV Y Runway 5R",
+          "runway": "05R",
+          "tower_controller": "IND_W_TWR",
+          "type": "RNAV",
+          "waypoints": [
+            "KELLY/a5000 IIUUU/a5000/s210 BIGRD/a4000 BGTEN/a3000 HUUPS/a2100 _IND_05R/a791"
+          ]
+        },
+        "I14": {
+          "full_name": "ILS Runway 14",
+          "runway": "14",
+          "tower_controller": "IND_E_TWR",
+          "type": "ILS",
+          "waypoints": [
+            "JELLS/a4000 KEADY/a4000/s210 RSADE/a2400 YEAAR/a1320 _IND_14/a796"
+          ]
+        },
+        "R14": {
+          "full_name": "RNAV Y Runway 14",
+          "runway": "14",
+          "tower_controller": "IND_E_TWR",
+          "type": "RNAV",
+          "waypoints": [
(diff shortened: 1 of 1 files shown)


sim.go
```diff
@@ -884,6 +939,28 @@ func newWorld(ssc NewSimConfiguration, s *Sim, sg *ScenarioGroup, sc *Scenario)
...
+				wind += fmt.Sprintf("G%02d", gst)
+			}
+		}
+		wind += "KT"
+
+		// Just provide the stuff that the STARS display shows
+		w.METAR[icao] = &METAR{
+			AirportICAO: icao,
+			Wind:        wind,
+			Altimeter:   fmt.Sprintf("A%d", alt-2+rand.Intn(4)),
```
mmp, line 961: It would be nice if `getweather` returned the altimeter as well since it is right there in the JSON from aviationweather.gov. How about returning an instance of `MetarData` since you've got that handy anyway?

sim.go
```diff
@@ -347,8 +349,13 @@ func (c *NewSimConfiguration) SetScenario(groupName, scenarioName string) {
 	}
 	c.ScenarioName = scenarioName
 }
-
+var WindDirection, WindSpeed, WindGust int32
+var countingNumber int
 func (c *NewSimConfiguration) DrawUI() bool {
+	if countingNumber == 0 {
+		countingNumber += 1
+		go getWind(c)
```
mmp, line 357: This ends up displaying the wind for whatever airport happens to be the first in `c.Scenario.ArrivalRunways`, which seems semi-arbitrary. It might be worth using `sg.PrimaryAirport` for the winds shown in the UI?

sim.go
```diff
@@ -347,8 +349,13 @@ func (c *NewSimConfiguration) SetScenario(groupName, scenarioName string) {
 	}
 	c.ScenarioName = scenarioName
 }
-
+var WindDirection, WindSpeed, WindGust int32
```
mmp, line 352: It's nice to avoid global variables in general and one issue here is that it doesn't handle multiple airports well (so the wind shown on the STARS scope will be the same for all airports for example.)

Here is a proposal for a different structure that would handle that better:
first, to update the `getweather` API to be something like
```
func GetAirportWeather(icao string) (MetarData, bool)
```
where `GetAirportWeather` a) queues a request for the airport's weather if it doesn't have it and hasn't yet queried the website (returning an empty struct and false), b) returns the weather (and true) if it has come back from the website, and c) does nothing if it has already queued up a request (also returning an empty struct and false in this case.)

Then the `DrawUI` code can call that function as much as it wants, each time it's called and then just use the info that comes back if it's available.

The nice thing about colly then is that it can do the http request in the background for you, so you don't need to do anything with goroutines here in sim.go. If you initialize the collector like:
```
		c = colly.NewCollector(
			colly.UserAgent("Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"),
			colly.Async(true))
```
then when you call c.Visit(), it will always return immediately. Further, if you give it the same URL more than once, it returns `colly.ErrAlreadyVisited`.  (And then as before the func you pass `OnResponse` will be called whenever the response comes back in the future.) (Note that you'd want the `colly.Collector` to be a global in `getweather` since it needs to remember which URLs it's seen..

So then if `getweather` maintains something like `map[string]MetarData`, then when `GetAirportWeather` is called it can return the weather if it's in the map and otherwise queue a request. It can ignore it if colly returns `ErrAlreadyVisited`, and return false in any case if the weather isn't in the map. (And then the `c.OnResopnse` func updates the map...)

Centralizing the logic for "has this been fetched, am I waiting for something to come back, etc" over in `getweather` also lets you simplify / remove things like your `countingNumber` here...

sim.go
```diff
@@ -495,16 +502,45 @@ func (c *NewSimConfiguration) DrawUI() bool {
...
+
+			}
+			
+			
 			if wind.Gust > wind.Speed {
+				
 				imgui.Text(fmt.Sprintf("%03d at %d gust %d", wind.Direction, wind.Speed, wind.Gust))
 			} else {
-				imgui.Text(fmt.Sprintf("%03d at %d", wind.Direction, wind.Speed))
+				imgui.Text(fmt.Sprintf("%03d at %d", WindDirection, wind.Speed))
```
mmp, line 543: should `wind.Speed` be `WindSpeed`? (Possibly a moot comment depending on possible API changes / removing the globals...)

sim.go
```diff
@@ -884,6 +939,28 @@ func newWorld(ssc NewSimConfiguration, s *Sim, sg *ScenarioGroup, sc *Scenario)
 		}
 	}
 
+	realMETAR := func(icao string) {
```
mmp, line 942: One issue with the suggestion above about changing the API is that the first time it sees an airport it will return false and then at this point it wouldn't have the information ready and you wouldn't be able to get the METAR. One option would be to pre-fetch the METAR for the departure and arrival airports in `DrawUI()` and just ignore the value returned so that by the time this code ran the METAR would likely be available.

That is admittedly a little ugly but I think the alternative of for example a `getweather` call that blocked until it had a response from the weather server wouldn't be a good idea in case the network was down. One nice thing about `colly` is that is handled by it (with the async option), in which case it will just never call your callback function...

mmp (commented):
Good start and forgive me for having lots of comments! (And sorry for not digging into this earlier..)

PR #24, opened 2024-01-19
Improved departure message

checkandmate1:
Small improvement to the departure message of departing aircraft. Includes the altitude they are climbing too.


mmp:
Looks good!

PR #26, opened 2024-02-20
Live Weather

checkandmate1:
Added live weather to _vice_. Few things:

- _vice_ will generate a global map; `airportWind`, which will map the primary airport (string) to the wind struct so that when the same airport is called again, it refers to the map instead of making another request. 
- The AirportWeatherData API works differently now. Utilizes colly asynchronously and returns the entire weather struct instead of just the wind so that the altimeter can also be correct.
- If KAAC is selected or _vice_ cannot find a connection, _vice_ will use the `uiStartDisable`/`uiEndDisable` functions so that the checkbox for toggling live weather cannot be selected.


checkandmate1:
I'm going to change this to the CWT branch

PR #29, opened 2024-02-24
CWT + Live Weather + Bug fixes

checkandmate1:
Live Weather:

 - vice will generate a global map; airportWind, which will map the primary airport (string) to the wind struct so that when the same airport is called again, it refers to the map instead of making another request.
- The AirportWeatherData API works differently now. Utilizes colly asynchronously and returns the entire weather struct instead of just the wind so that the altimeter can also be correct.
- If KAAC is selected or vice cannot find a connection, vice will use the uiStartDisable/uiEndDisable functions so that the checkbox for toggling live weather cannot be selected.

CWT:

 - CWT (RECAT Category) is determined via the openscope aircraft file
     * There is a category "H" in the openscope file, I'm not entirely sure why that is, so if an aircraft is category H *vice* will treat it as a category F aircraft
     * If an aircraft RECAT category is unknown, it will display "NOWGT" (This shouldn't happen because all aircraft have RECAT categories, but it is there)
- CWT/ RECAT will be used if it is specified in the facility config file with the bool `recat_facility` (eg. `"recat_facility": true`)
- If it is specified in the facility configuration file, ATPA separation minima will be updated to show the proper wake turbulence separation
- CWT will show on partial ghost targets (Figure 6-2: CRDA Illustration Showing Ghost Tracks in Stagger Mode)

Bug Fixes:

- Altitude filters are fixed. FDBs are shown for associated tracks under these conditions (STARS Manual 2.12):
   * track is controlled at this TCW/TDW
   * track (unowned) has been "Quick Looked" at this TCW/TDW track meets adapted criteria and is in an enabled Quicklook region
   * track is involved in a handoff with this TCW/TDW • track is in "Point out" to this TCW/TDW
  *  track is squawking a Special Condition beacon code (Lost Link, General Emergency, Hijack, Radio Communications Failure, or Military Intercept)
  *   track has Minimum Safe Altitude Warning (MSAW) or Conflict Alert (CA) indicators.
  *   track is displaying an Automated Terminal Proximity Alert (ATPA) Caution or Alert Cone
  *  track is in an ATPA Warning or Alert condition, the display of ATPA In-trail Distance is enabled, and the TCP is adapted to display ATPA In-trail Warning and Alerts for the ATPA Approach Volume.


stars.go
```diff
@@ -5402,15 +5541,20 @@ func (sp *STARSPane) formatDatablocks(ctx *PaneContext, ac *Aircraft) []STARSDat
 		} else if sp.isOverflight(ctx, ac) {
 			field3 += "E"
 		}
-
-		actype := ac.FlightPlan.TypeWithoutSuffix()
+		if ctx.world.RECAT {
+			cat := getRecatCategory(ac)
+			field3 = cat
```
mmp, line 5546: Minor, but should this be `field3 += cat`, similar to the non-RECAT stuff below? (i.e. if it's VFR or an overflight I think we still want the V/E in there?)

aviation.go
```diff
@@ -1215,6 +1215,11 @@ type AircraftPerformance struct {
 		Accelerate float32 `json:"accelerate"` // kts / 2 seconds
 		Decelerate float32 `json:"decelerate"`
 	} `json:"rate"`
+	Category struct {
```
mmp, line 1218: please run `gofmt -w aviation.go` (and similarly on other modified files) to reindent / align  the types inside structs, etc.

sim.go
```diff
@@ -613,6 +653,29 @@ func (c *NewSimConfiguration) DrawUI() bool {
 	return false
 }
 
+func getWind(c *NewSimConfiguration) Wind {
+	airport := c.Scenario.PrimaryAirport
+
+	weather, errorss := getweather.GetWeather(airport)
```
mmp, line 659: `errorrss` -> `err` for consistency

sim.go
```diff
@@ -613,6 +653,29 @@ func (c *NewSimConfiguration) DrawUI() bool {
...
+
+	if len(errorss) > 0 {
+		fmt.Println("Error. ", errorss)
```
mmp, line 662: This should probably be a `lg.Info`.

stars.go
```diff
@@ -4779,12 +4781,14 @@ func (sp *STARSPane) drawGhosts(ghosts []*GhostAircraft, ctx *PaneContext, trans
...
 			datablockText = fmt.Sprintf("%02d", (ghost.Groundspeed+5)/10)
-
-			fp := ctx.world.Aircraft[ghost.Callsign].FlightPlan
-			fields := strings.Split(fp.AircraftType, "/")
-			if len(fields) > 1 && (fields[0] == "H" || fields[0] == "J" || fields[0] == "S") {
-				datablockText += fields[0]
-			}
+			ac := ctx.world.Aircraft[ghost.Callsign]
+			recat := getRecatCategory(ac)
+			// fp := ctx.world.Aircraft[ghost.Callsign].FlightPlan
```
mmp, line 4786: Was it intended to comment these lines out? I assume you want to be testing world.RECAT here and then formatting the ghost datablock accordingly?

stars.go
```diff
@@ -5280,15 +5290,143 @@ func (sp *STARSPane) checkInTrailSeparation(back, front *Aircraft) {
...
+		return "C"
+	case 'B':
+		return "B"
+	case 'A':
+		return "A"
+	case 'H':
+		return "F"
+	default:
+		lg.Errorf("%s: unexpected weight class \"%c\"", ac.Callsign, wc[0])
+		fmt.Printf("%s: unexpected weight class \"%c\"\n", ac.Callsign, wc[0])
```
mmp, line 5334: Can delete this--iIn general, we don't want to be printing things in the normal course of operation but instead sending them to the logs. Printf is definitely useful when initially getting things going, though...

stars.go
```diff
@@ -5280,15 +5290,143 @@ func (sp *STARSPane) checkInTrailSeparation(back, front *Aircraft) {
...
+		return 4
+	case 'A':
+		return 5
+	default:
+		lg.Errorf("%s: unexpected weight class \"%c\"", ac.Callsign, wc[0])
+		return 1
+	}
+}
+
+// We don't have the info we need for RECAT in the openscope
```
mmp, line 5371: Please update the comment so it's up to date...

mmp (commented):
Looks good overall; really nice additions to have. Here is an initial round of comments/suggestions. I'l spend more time on the weather stuff tomorrow.

stars.go
checkandmate1: This makes more sense, thanks

main.go
```diff
@@ -46,6 +46,7 @@ var (
 	newWorldChan chan *World
 	localServer  *SimServer
 	remoteServer *SimServer
+	airportWind  map[string]Wind
```
mmp, line 49: Is this ever initialized with a `make(map[string]Wind)`? Also, I don't think anything is ever assigned to it.

sim.go
```diff
@@ -496,17 +500,55 @@ func (c *NewSimConfiguration) DrawUI() bool {
...
+			}
+			uiEndDisable(!validAirport)
 			imgui.TableNextColumn()
-			wind := c.Scenario.Wind
+			var wind Wind
+			if !c.LiveWeather {
+				wind = c.Scenario.Wind
+			} else {
+				empty := Wind{}
+				if airportWind[c.Scenario.PrimaryAirport] != empty {
```
mmp, line 526: A more idiomatic way to write it would be to remove `empty` and do:

```
var ok bool
if wind, ok = airportWind[c.Scenario.PrimaryAirport]; !ok {
  // fetch wind...
}
```

sim.go
```diff
@@ -496,17 +500,55 @@ func (c *NewSimConfiguration) DrawUI() bool {
...
+					wind = airportWind[c.Scenario.PrimaryAirport]
+				} else {
+					liveWind := make(chan Wind)
+					go func(liveWinds chan Wind) {
```
mmp, line 530: The goroutine here ends up not having any effect: since `liveWind` is unbuffered, `wind = <-liveWind` will stall until the goroutine runs and returns a value through the `liveWinds` channel. (And so in turn, the weather will be fetched every time the UI is redrawn--tens of times a second, and the UI drawing will be slowed down waiting for the network round trip, and may hang if the network is down, etc.) See below for a proposal to adapt `getWind()` to work around this...

sim.go
```diff
@@ -613,6 +655,29 @@ func (c *NewSimConfiguration) DrawUI() bool {
 	return false
 }
 
+func getWind(c *NewSimConfiguration) Wind {
```
mmp, line 658: It might make more sense to pass in `c.Scenario.PrimaryAirport` and to have getWind just take a string parameter, since it doesn't access anything else in `c` or `c.Scenario`.

sim.go
```diff
@@ -613,6 +655,29 @@ func (c *NewSimConfiguration) DrawUI() bool {
...
+	airport := c.Scenario.PrimaryAirport
+
+	weather, err := getweather.GetWeather(airport)
```
mmp, line 661: I think it's important that the main flow of the program running never be held up by waiting for something over the network, not only because that may make it less responsible but also that it may make vice hang if a network request stalls.

So, I propose that `getWind` be rewritten to return `(Wind, bool)`, where the bool indicates whether it has the wind, and where it returns immediately if the wind isn't available. The first time wind is requested for an airport the network request can be initiated and then once it comes back the wind can start being returned.

So if you add a global (all of this is untested)
```
var windRequests map[string]chan getweather.[]MetarData
```
then the flow is basically: 1. is the airport's wind already cached in `airportWind`? If so, return that wind, true. 2. has the weather been requested already? If so, return no wind, false:
```
if _, ok := windRequests[airport]; ok {
   // it's been requested but we don't have it yet
   return Wind{}, false
}
```
3. otherwise request it:
```
c := make(chan []airportweather.MetarData)
windRequests[airport] = c
go func() {
    weather, err := getweather.GetWeather(airport)
    // handle errors
   c <- weather
}()
return Wind{}, false
```

Then (probably at the start of getWind) you would want to check all of the `chan`s to see if they have gotten anything back:
```
for airport, ch := range windRequests {
    select {
        case w := <- ch:
            airportWind[airport] = w
            delete(windRequests, airport)
        default:
            // no wind yet
    }
}
```

mmp (commented):
A bunch of small things and then a bigger comment/suggestion about making `getWind` be non-blocking.

PR #30, opened 2024-03-06
STARS v2: Part 1

checkandmate1:
Lots of new features. Probably should be a beta release because there may be a few bugs that when running a local server and two instances of _vice_ at the same time on a MacBook were not found:

- Specifying positions
     - Same-sector positions can be specified with one letter (eg. `A` to handoff to `4A` working `4P`
     - Interfacility positions must be specified with the delta (∆) symbol, the facility identifier, and then the sector ID
     - If specified in the facility configuration file, `C` can be used as a handoff to the appropriate center controller
     - If specified in the facility configuration file, interfacility handoffs can be specified with the delta symbol and the facility ID
- Handoffs
    -  More realism with the syntax required for certain handoffs
- Misc. STARS Commands
     - `**` will Force QL the target to the user, if they have control of the track and it is allowed by the STARS facility (specified in the facility configuration file)
     - `** + SECTORID` will force QL a track to that intrafacility position
     - `MULTI FUNC, O` will display the last 20 TCPs that have accepted a point out
     - QL a TCP without needing to use `MULI FUNC, Q` (yay!)
     - Scratchpads may now be set with `SLEW` instead of `MULTI FUNC, Y` The Length of scratchpads defaults at 3, but can be extended to four if allowed by the STARS facility. (Specified in the facility configuration file)
 - Facility configuration file changes
     - The `Controller` struct now needs a `facility_identifier` JSON field
     - `NewSimConfigration` now needs the `stars_adaptation` JSON field containing: a bool: `force_ql_self`, a [2]bool: `scratchpad_rules` bool[0] is for the primary scratchpad, bool[1] is for the secondary scratchpad, and the `airspace_awareness` struct. Inside of this struct there is a []string: `fix` which is a slice of fixes that an aircraft could be filed over. A string `altitude_range` which is specified as `100450` for 10,000 to 45,000. A string`receiving_controller` which is the controller that will receive this. Finally the book: `to_center` which is pretty self-explanatory 
 - TODO List
     - ~~Redirecting Handoffs~~
     - `MULTI FUNC ZZ`
     - Unassociated tracks
     - ~~LDB vs PDB~~
     - Needing to specify the BCN or ACID to initiate control
     - Editing who gets what during multiplayer sessions. (So that CAMRN doesn't have to work Yardley if HAARP is online, the users can decide)
     - DCB size is not one set size but depends on the screen size (My DCB takes up the entire top of my screen, while Jud's for example is 1/3 - 1/2 his screen size
     - ~~Global Leader Lines~~

- Fixes
   - Fixed aircraft going direct after being handed off to a real controller

Let me know if anything should be changed (probably a lot that is going to be needed too)!


checkandmate1:
Looks like I was able to add in redirecting handoffs, and I feel even stronger about this being a beta, considering people will find a way to break it.

mmp:
Great stuff. I have a few small comments / questions that I will try to send today. Just to simplify it a bit I will also start integrating some of this into the master branch via individual per-feature commits.

eventstream.go
```diff
@@ -193,6 +193,8 @@ const (
 	IdentEvent
 	HandoffControllEvent
 	NumEventTypes
+	ForceQLEvent
```
mmp, line 196: `NumEventTypes` should stay the last value here so that it has the right value.

eventstream.go
```diff
@@ -193,6 +193,8 @@ const (
...
+	AcknowledgedForceQLEvent
 )
 
 func (t EventType) String() string {
```
mmp, line 200: Please update this for the new event types; it will likely panic if called for either of them currently.

aviation.go
```diff
@@ -237,6 +234,8 @@ type Controller struct {
 	SectorId  string    `json:"sector_id"`  // e.g. N56, 2J, ...
 	Scope     string    `json:"scope_char"` // For tracked a/c on the scope--e.g., T
 	IsHuman   bool      // Not provided in scenario JSON
+	FacilityIdentifier string `json:"facility_identifier"` // For example the "N" in "N4P" showing the N90 TRACON
```
mmp, line 237: Could you rename the JSON "facility_identifier" to be "facility_id"--just to be consistent with the "id" in "sector_id"

stars.go
```diff
@@ -5481,9 +6083,21 @@ func (sp *STARSPane) formatDatablocks(ctx *PaneContext, ac *Aircraft) []STARSDat
 			field8 = " PO"
 		} else if id, ok := sp.OutboundPointOuts[ac.Callsign]; ok {
 			field8 = " PO" + id
-		} else if _, ok := sp.RejectedPointOuts[ac.Callsign]; ok {
+		} else if _, ok := sp.RejectedPointOuts[ac.Callsign]; ok && time.Now().Second()&1 == 0 {
```
mmp, line 6086: Don't look at the time in `formatDatablocks` for things that are flashing/alternating but instead have it return multiple datablocks, one with each variant (on/off, etc.) The reason it always returns all datablock variants is so that the code that draws them can compute a bounding box of the datablock text that is consistent across flashing/alternating stuff. In that way, the datablock doesn't potentially shift to the left and right as it alternates. (Depends on how the datablock is positioned with respect to the track.)

(Applies a few places here)

stars.go
```diff
@@ -5607,28 +6227,48 @@ func (sp *STARSPane) datablockColor(w *World, ac *Aircraft) (color RGB, brightne
...
+		}
+	}
 
 	// Handle cases where it should flash
 	now := time.Now()
 	if now.Second()&1 == 0 { // one second cycle
 		_, pointOut := sp.InboundPointOuts[ac.Callsign]
-		if state.Ident() || // ident
-			ac.HandoffTrackController == w.Callsign || // handing off to us
+		if (ac.HandoffTrackController == w.Callsign && !strings.Contains(fmt.Sprintf("%v", ac.RedirectedHandoff.Redirector), userController.SectorId)) || // handing off to us
```
mmp, line 6242: To save the cost/round-trip of converting `Redirector` to a string, you can replace `!strings.Contains(fmt.Sprintf("%v", ac.RedirectedHandoff.Redirector, userController.SectorId))`  with `!slices.Contains(ac.RedirectedHandoff.Redirector, userController.SectorId)`

stars.go
```diff
@@ -5607,28 +6227,48 @@ func (sp *STARSPane) datablockColor(w *World, ac *Aircraft) (color RGB, brightne
...
 		color = STARSInboundPointOutColor
 	} else if state.IsSelected {
 		color = STARSSelectedAircraftColor
 	} else if ac.TrackingController == w.Callsign {
 		color = STARSTrackedAircraftColor
-	} else if ac.HandoffTrackController == w.Callsign {
+	} else if ac.RedirectedHandoff.OrigionalOwner == userController.SectorId || ac.RedirectedHandoff.RedirectedTo == userController.SectorId {
+		color = STARSTrackedAircraftColor
+	} else if (ac.HandoffTrackController == w.Callsign &&
+		!strings.Contains(fmt.Sprintf("%v", ac.RedirectedHandoff.Redirector), userController.SectorId)) ||
```
mmp, line 6270: (same comment as above)

mmp (commented):
A handful of small things. Looks great overall.

eventstream.go
```diff
@@ -193,6 +193,8 @@ const (
 	IdentEvent
 	HandoffControllEvent
 	NumEventTypes
+	ForceQLEvent
+	AcknowledgedForceQLEvent
```
mmp, line 197: Actually, you might just remove these two new event types (and the code that posts them in `sim.go`, since there isn't any code that's listening for them. The event stream stuff is mostly used for propagating information that isn't really explicitly represented anywhere whereas the force QL stuff is right there in `Aircraft`, so `STARSPane` can do its thing with it directly.

stars.go
```diff
@@ -1584,8 +1593,14 @@ func (sp *STARSPane) processKeyboardInput(ctx *PaneContext) {
 	for key := range ctx.keyboard.Pressed {
 		switch key {
 		case KeyBackspace:
+
 			if len(sp.previewAreaInput) > 0 {
-				sp.previewAreaInput = sp.previewAreaInput[:len(sp.previewAreaInput)-1]
+				if len(sp.previewAreaInput) >= 2 && string(sp.previewAreaInput[len(sp.previewAreaInput)-2]) == "Â" {
```
mmp, line 1598: I think Â is \u00c2? Let's make a const value to use here and other places where \u00c2 is used, e.g.

const STARSTriangleCharacter = "\u00C2"

somewhere up toward the top of the file.

stars.go
```diff
@@ -1558,7 +1567,7 @@ func (sp *STARSPane) processKeyboardInput(ctx *PaneContext) {
 		input = input[1:]
 	}
 	sp.previewAreaInput += input
-
+	sp.previewAreaInput = strings.Replace(sp.previewAreaInput, "`", "\u008A", -1)
```
mmp, line 1570: Should \u008A be \u00C2 (aka STARSTriangleCharacter or whatever) or is this something else?

stars.go
```diff
@@ -1858,6 +1872,177 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
+							sp.forceQL(ctx, aircraft.Callsign, control)
+							status.clear = true
+						}
+						return
+					}
+				}
+
+			} else {
+				for _, aircraft := range ctx.world.Aircraft {
+					if len(cmd) > 8 && cmd[:7] == aircraft.Callsign {
```
mmp, line 1904: You could eliminate the separate cases for each length (and the all of repeated code) by useing the `lookupCallsign` utility function defined way at the top of this function to do something like:

```
callsign, tcps, _ := strings.Cut(cmd, " ")
aircraft := lookupCallsign(callsign)
if aircraft != nil {
        for _, tcp := range strings.Split(tcps, " ") {
            // as before...
```

sim.go
```diff
@@ -2134,10 +2208,73 @@ func (s *Sim) CancelHandoff(token, callsign string) error {
...
+			octrl := s.World.GetController(controller)
+			s.eventStream.Post(Event{
+				Type:           AcknowledgedForceQLEvent,
+				FromController: ctrl.Callsign,
+				ToController:   octrl.Callsign,
+				Callsign:       ac.Callsign,
+			})
+			for i, s := range ac.ForceQLControllers {
+				if s == controller {
+					ac.ForceQLControllers = append(ac.ForceQLControllers[:i], ac.ForceQLControllers[i+1:]...)
```
mmp, line 2267: I'm not sure if this will work if the same controller happens to be in the list twice. There's a `FilterSlice` utility function in util.go that you could use in place of the `for` loop:
```
ac.ForceQLControllers = FilterSlice(ac.ForceQLControllers, func(qlController string) bool { return qlController != controller })
```

sim.go
```diff
@@ -2006,7 +2079,7 @@ func (s *Sim) HandoffTrack(token, callsign, controller string) error {
 		})
 }
 
-func (s *Sim) HandoffControl(token, callsign string) error {
+func (s *Sim) HandoffControl(token, callsign, controller string) error {
```
mmp, line 2082: I don't think controller needs to be passed in here. (It doesn't look like it's used?)

stars.go
```diff
@@ -1858,6 +1872,177 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
 			return
 		}
 
+		if len(cmd) > 5 && cmd[:2] == "**" { // Force QL
+			// Manual 6-69
+			cmd = cmd[2:]
+			if unicode.IsDigit(rune(cmd[0])) { //BCN code
+				bcn := cmd[:4]
+				for _, aircraft := range ctx.world.Aircraft {
+					if fmt.Sprintf("%v", aircraft.AssignedSquawk) == bcn {
```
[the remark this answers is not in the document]
mmp: (Actually, maybe lookupAircraft(), discussed in an upcoming comment, might be all you need here.)

stars.go
```diff
@@ -2366,6 +2616,31 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
 			}
 
 		case "O":
+			if len(cmd) > 2 {
+				specifiedAircraft := cmd
```
mmp, line 2620: It looks like this only looks up by beacon code, but I think the command takes either ACID, beacon code, or line number. There's a `lookupAircraft()` utility function at the top of this function that handles all of those cases, so you could just write something like this so that all three ways work.

```
aircraft := lookupAircraft(cmd)
if aircraft == nil {
  // etc.
```

stars.go
```diff
@@ -3119,6 +3574,51 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
+					if tcps[0] == "ALL" {
+						// Force QL for all TCP
+						// Find user fac
+						var fac string
+						for _, control := range ctx.world.Controllers {
+							if control.Callsign == ctx.world.Callsign {
+								fac = control.FacilityIdentifier
+							}
+						}
+						for _, control := range ctx.world.Controllers {
```
mmp, line 3602: I think you can merge the two `for` loops and eliminate the `fac` variable by having a single loop with the `if` test `if control.Callsign == ctx.world.Callsign && !control.ERAMFacility`

stars.go
```diff
@@ -3119,6 +3574,51 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
+							if !control.ERAMFacility && control.FacilityIdentifier == fac {
+								sp.forceQL(ctx, ac.Callsign, control.SectorId)
+							}
+						}
+					}
+					for _, tcp := range tcps {
+						ok, control := sameFacility(ctx, tcp, ac.Callsign)
+						if !ok {
+							// Do something idk
```
mmp, line 3611: Maybe `status.err = GetSTARSError(ErrSTARSIllegalPosition)`? But definitely don't clear the status and return since otherwise the TCPs after the bogus one will just be ignored and no error will be issued.

stars.go
```diff
@@ -4618,10 +5200,16 @@ func (sp *STARSPane) datablockType(w *World, ac *Aircraft) DatablockType {
...
 	}
 
-	if ac.TrackingController == w.Callsign || ac.ControllingController == w.Callsign {
+	if ac.TrackingController == w.Callsign || (ac.ControllingController == w.Callsign) {
 		// it's under our control
 		dt = FullDatablock
 	}
+	me := w.GetController(w.Callsign)
+	for _, control := range ac.ForceQLControllers {
+		if control == me.Callsign {
```
mmp, line 5209: It should work to skip the `w.GetController` call and just do `if control == w.Callsign`

mmp (commented):
Here are some comments/suggestions related to ForceQL. (I'm planning to merge that in to the master branch next.)

stars.go
checkandmate1: `ac.RedirectedHandoff.Redirector` and `ac.RedirectedHandoff.RedirectedTo` take in the sector IDs for the controllers instead of the callsign, so the call will have to be made somewhere. I might be smarter to change it to a callsign and just get the controller from that callsign

stars.go
checkandmate1: Ugh yup, I put this there while I looked into the docs and tried to find the correct error message, but I probably got sidetracked with another feature. Looks like ILL POS is correct.

stars.go
checkandmate1: Oh great, I didn't realize that this was a function.

stars.go
checkandmate1: It's really weird. \u008A is the ∆ symbol, but I think when you were mentioning how its two bytes, one of those bytes is \u00c2. When I use \u00c2 it comes out as a question mark

stars.go
checkandmate1: FYI, I changed `lookupCallsign()` to ` lookupAircraft()` because if `lookupCallsign()` fails then it will just return the entered callsign

mmp:
When you have a chance, would you mind running `gofmt -w *.go` in your vice directory and commiting and pushing the commit to this branch? I'm ready to pull the force-QL and facility_id / eram_facility changes into the main branch and it'll save future merge main to have those formatted consistently when they first go in.

stars.go
```diff
@@ -2890,20 +3027,182 @@ func (sp *STARSPane) acceptHandoff(ctx *PaneContext, callsign string) {
...
+		for _, fix := range rules.Fix {
+			if strings.Contains(ac.FlightPlan.Route, fix) {
+				if rules.AltitudeRange == "" {
+					return rules.ReceivingController, nil
+				} else {
+					alt, err := breakAltitude(rules.AltitudeRange)
+					if err != nil {
+						return "", errors.New(fmt.Sprintf("Error breaking %v: %v", rules.AltitudeRange, err))
+					}
+					if ac.FlightPlan.Altitude >= alt[0] && ac.FlightPlan.Altitude <= alt[1] {
```
mmp, line 3122: Should this check be based on the flight plan altitude or the aircraft's current altitude? (I could imagine it either way so just wanted to confirm.)

mmp (commented):
Few more small things.

stars.go
checkandmate1: I was talking to an IRL PHL TRACON controller and he said it's cruise altitude. I would imagine that they flash to the center as soon as the aircraft appears on the scope, so 500 feet MSL wont mean much

mmp: Got it-totally makes sense.

stars.go
```diff
@@ -2450,6 +2482,27 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
+					return
+				} else if aircraft.TrackingController == "" {
+					status.err = GetSTARSError(ErrSTARSIllegalTrack)
+				}
+
+				if len(aircraft.PointOutHistory) == 0 {
+					status.clear = true
+					return
+				}
+				var out string
```
mmp, line 2498: This could also be written just as `status.output = strings.Join(aircraft.PointOutHistory, " ")` to save a few lines of code. (I'll do this in a followon commit.)

stars.go
```diff
@@ -3177,6 +3293,20 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 						}
 					}
 				} else if ac.HandoffTrackController == ctx.world.Callsign {
+					if len(cmd) > 0 { // Redirecting handoff
```
mmp, line 3296: `len(cmd) > 0` won't ever be true here since there's an `if cmd == ""` above it. Should this code live somewhere else or is it redundant?

stars.go
```diff
@@ -3458,19 +3608,82 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
+					status.clear = true
+					sp.pointOut(ctx, ac.Callsign, control)
+					return
+				}
+
+			} else if len(cmd) > 0 {
+				// Handoffs
+				user := ctx.world.GetController(ctx.world.Callsign)
+				if ac.HandoffTrackController == user.Callsign || ac.RedirectedHandoff.RedirectedTo == user.SectorId { // Redirect
+					if cmd[0] == '\u00C2' {
```
mmp, line 3641: It looks like it's handled here so I will plan to go ahead and delete the above code when I merge the redirect handoff stuff.

stars.go
```diff
@@ -3458,19 +3608,82 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
+						cmd = cmd[2:]
```
mmp, line 3642: Also, is removing the leading ∆ necessary/desirable here? It looks like `sameFacility` has logic to do that and in particular only allows the ∆ if it's a different facility, while if it's removed here I think it requires the same facility? (I may be misreading the logic though.)

stars.go
```diff
@@ -3174,9 +3282,16 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 						}
 					}
 				} else if ac.HandoffTrackController == ctx.world.Callsign {
-					// Accept inbound h/o
+					// Redirecting handoff
```
mmp, line 3285: I don't think this case will ever hit because there's one above that already covers `ac.HandoffTrackController == ctx.world.Callsign`. So I will delete this in the merge but heads up if it's functionality that should still be there but under different logic.

stars.go
```diff
@@ -1531,6 +1534,10 @@ func (sp *STARSPane) updateRadarTracks(w *World) {
 
 	for callsign, state := range sp.Aircraft {
 		ac, ok := w.Aircraft[callsign]
+		if distance2f(ac.Position(), w.Center) > 1.66 && ac.IsDeparture() {
```
mmp, line 1537: Was this added to fix something? There's code in sim.go that culls aircraft that are > 250nm away (line 1546). If that distance needs to be tightened up then it should probably live in `Sim` rather than `STARSPane`.

(Also, just so you know, there's an `nmdistance2ll` function that returns the distance in nautical miles between two lat-long coordinates.)

stars.go
mmp: Yeah, I was wondering about that. I think we should be consistent one way or the other just to reduce the chance of bugs from getting the two cases mixed up. I'd suggest sticking with callsigns (in part for easier debugging when there are bugs), but am open to sector ids if there's a reason to stick with them.

stars.go
```diff
@@ -3133,22 +3172,32 @@ func sameFacility(ctx *PaneContext, controller, callsign string) (bool, string)
...
-			if userController.FacilityIdentifier != receivingController.FacilityIdentifier {
+			if receivingController.FacilityIdentifier != "" && string(controller[0]) == receivingController.FacilityIdentifier{
+				state := sp.Aircraft[callsign]
+				state.LastKnownHandoff = receivingController.Scope
 				return true, receivingController.SectorId
 			}
 
-		}
+		} 
+			for _, control := range ctx.world.Controllers {
```
mmp, line 3193: Please give this (and everything) a run through `gofmt -w` to harmonize indentation..

world.go
```diff
@@ -467,7 +477,24 @@ func (w *World) GetAircraft(callsign string) *Aircraft {
...
+
+func findAircraft(sample string, aircraft []*Aircraft) *Aircraft {
+
+	var final []*Aircraft
+	for _, icao := range aircraft {
+		if strings.Contains(icao.Callsign, sample) {
+			final = append(final, icao)
+		}
+	}
+	if len(final) > 1 || len(final) == 0 {
```
mmp, line 493: Could make the small simplification to the logic (and make it easier to follow) by:
```
if len(final) == 1 {
    return final[0]
} else {
    return nil
}

File: stars.go [hunk 27]
```diff
@@ -3349,10 +3398,28 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 					return
 				}
 				return
-			} else if dir, ok := numpadToDirection(cmd[0]); ok && len(cmd) == 1 {
-				state.LeaderLineDirection = dir
-				status.clear = true
-				return
+			} else if dir, ok := numpadToDirection(cmd[0]); ok && (len(cmd) == 1 || (len(cmd) == 2 && unicode.IsDigit(rune(cmd[1])))) {
```
mmp, line 3401: The test `ok && ...` is a bit long and then is partially repeated in the code below. What about:
```
} else if dir, ok := numpadToDirection(cmd[0]); ok && len(cmd) == 1 {
   // len 1 case below
} else if ok && len(cmd) == 2 {
  if !unicode.IsDigit(rune(cmd[1]) || cmd[0] != cmd[1] || strings.Contains(cmd, "0") {
     // error
  }
  // as before for global
}

File: aircraft.go [hunk 28]
```diff
@@ -21,6 +22,7 @@ type Aircraft struct {
 	FlightPlan          *FlightPlan
 	ForceQLControllers  []string
 	PointOutHistory     []string
+	POFlashingEndTime   time.Time
```
mmp, line 25: I think it would be better for this to live in STARSAircraftState since it's more of a per-user concern and it's related to display on the STARS scope, rather than something all users may care about. (Admittedly things like ForceQLControllers blur that line, but this is one that can pretty easily live entirely in STARS-land.)

So if you move this to `STARSAircraftState` and then around line 1290 in stars.go where it processes a point out event, you can initialize `POFlashingEndTime` there. (And then just rewire how it's used in `formatDatablocks`

stars.go
```diff
@@ -3680,9 +3746,25 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 				if len(cmd) == 1 {
 					if dir, ok := numpadToDirection(cmd[0]); ok {
 						state.LeaderLineDirection = dir
+						state.ChosenLeaderLine = dir
 						status.clear = true
 						return
 					}
+				} else if len(cmd) == 2 { // Global leader lines
```
mmp, line 3753: Actually, I'd suggest putting all this leader line stuff into a separate function since it's a lot of duplicated code now. (Admittedly it was duplicate code before but now it's a bunch more code.)  So both this code and the code above could be turned into a method with a signature something like:
```
func (sp *STARSPane) setLeaderLine(state *STARSAIrcraftState, cmd string) (bool, error)
```
that's called something like `status.clear, status.err = sp.setLeaderLine(...)` (as is done for a few other methods.)

stars.go
```diff
@@ -5869,39 +5914,50 @@ func (sp *STARSPane) formatDatablocks(ctx *PaneContext, ac *Aircraft) []STARSDat
...
 
-	var errs []string
-	if state.MSAW && !state.InhibitMSAW && !state.DisableMSAW && !ps.DisableMSAW {
-		errs = append(errs, "LA")
+	if state.ChosenLeaderLine == nil {
+		state.LeaderLineDirection = ac.GlobalLinePosition
+	} else {
+		state.LeaderLineDirection = state.ChosenLeaderLine
+	}
+	index := slices.Index(state.Warnings, state.SPCOverride)
```
mmp, line 5922: Does Warnings need to be persistent? It looks like it's only ever used in this function. Also, likely related, if I put aircraft into a CA situation, the datablock shows "CA/CA/CA/CA/CA/..."; it looks like this is never reset and only appended to.

stars.go
```diff
@@ -3441,13 +3476,16 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
 				}
-
-				if state.DatablockType != FullDatablock {
-					state.DatablockType = FullDatablock
+				db := sp.datablockType(ctx.world, ac)
+				if db == LimitedDatablock && time.Until(state.FullLDB) <= 0 {
+					state.FullLDB = time.Now().Add(5 * time.Second)
 					// do not collapse datablock if user is tracking the aircraft
-				} else if ac.TrackingController != ctx.world.Callsign {
+				} else if db == FullDatablock {
```
mmp, line 3483: Double-checking this change: is it allowed to convert a FDB for a PDB for aircraft under our control?

(Specifically asking since it was changed to be the current logic in PR #4, which came from MK, who said the previous behavior (which is the same as your change) was wrong.)

checkandmate1: This was for collapsing FDBs that can collapse, but forgot to add the logic for the ones that shouldn't... Thanks

stars.go
```diff
@@ -2351,8 +2381,6 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
 					// Tracked by other controllers
 					ps.OtherControllerLeaderLineDirection = dir
 					status.clear = true
-				} else {
```
mmp, line 2354: Was deleting this intentional? Now if I enter `[multifunc]7AA[enter]` the preview area text remains there and no error is issued.

stars.go
```diff
@@ -2060,17 +2065,42 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
 			ok, control := sp.calculateController(ctx, cmd, "")
 			if !ok {
 				status.err = GetSTARSError(ErrSTARSIllegalPosition)
 				return
 			}
 			for _, controler := range ctx.world.Controllers {
-				if controler.SectorId == control {
-					positions, input, err := parseQuickLookPositions(ctx.world, cmd)
+				if controler.Callsign == control {
+					c := ctx.world.GetController(control)
```
mmp, line 2096: I think the call to `world.GetController()` here is unnecessary since the `for` loop is over `world.Controllers`. So for the merge I am simplifying this to:
```
for _, c := range ctx.world.Controllers {
    if c.Callsign == control {
       positions, input, err := ...
       // etc.
```
Let me know if I'm missing something!

stars.go
mmp: I'm still not following why `state.Warnings` needs to be persistent versus having `var errs []string` local in this function like before...

stars.go
```diff
@@ -2060,17 +2065,42 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
+					ps.QuickLookAllIsPlus = true
+				} else {
+					ps.QuickLookAll = !ps.QuickLookAll
+					ps.QuickLookAllIsPlus = false
+					ps.QuickLookPositions = nil
+				}
+				status.clear = true
+				return
+			}
 			ok, control := sp.calculateController(ctx, cmd, "")
```
mmp, line 2089: This is from a previous merge so I'm going to keep going with getting all this stuff merged, but it looks like we've lost the ability to QL multiple sector ids. Previously you could enter `[MULTIFUNC]1L 1V[enter]` and QL both 1L and 1V but now that gives ILL POS. (I will merge and file a bug to track this to fix afterward.)

checkandmate1: Oh, I never thought that quick-looking multiple-sector IDS was a thing. Should I fix it or will you fix it in the merge?

mmp: I left it as is in the merge (since this merge didn't introduce the regression) and filed a bug to track it. I can fix it or you can, either way.

checkandmate1: I can fix it after the merge. Sounds good.

stars.go
```diff
@@ -3126,6 +3127,9 @@ func (sp *STARSPane) handoffTrack(ctx *PaneContext, callsign string, controller
 // and route.
 func calculateAirspace(ctx *PaneContext, callsign string) (string, bool) {
 	ac := ctx.world.Aircraft[callsign]
+	if ac == nil {
+		return "no target", false
```
mmp, line 3131: Could you rewrite this to return an `error` with a custom error type for this case? In general, that's better than overloading the string return value, which in principle should be a controller id, with auxiliary information. Then it's more obvious in calling code what's going on versus checking against a random string.

(Same comment applies to returning "sp one" from `calculateController`, which should also probably be revisited.)

sim.go
```diff
@@ -2456,7 +2456,9 @@ type HeadingArgs struct {
 func (s *Sim) AssignHeading(hdg *HeadingArgs) error {
 	s.mu.Lock(s.lg)
 	defer s.mu.Unlock(s.lg)
-
+	if hdg.Heading%5 != 0 { // Just so that scratchpads like R11 can go in
+		return errors.New("Heading too specific")
```
catcarbon, line 2471: Not meant to block this PR but I think tweaks like this are telltale that we need to make simulator commands explicitly distinct from STARS commands.

stars.go
checkandmate1: I'm not at my computer right now, but from what I remember SPC override was a string that was entered into errs. So when it was replaced,the previous spc would be overridden by the new one. So basically, only one warning could be entered at a time.

sim.go
mmp: +1

stars.go
```diff
@@ -3175,76 +3207,86 @@ func singleScope(ctx *PaneContext, facilityIdentifier string) *Controller {
...
-			if controller != nil {
-				return true, controller.Callsign
+	if controller[0] == 'C' || (haveTrianglePrefix && lc == 3) {
+		if lc == 3 {
+			if control := singleScope(ctx, string(controller[2])); control != nil {
+				return control.Callsign, nil
+			}
+		}
+		if callsign == "" {
+			return "", errors.New("Cannot calculate airspace")
```
mmp, line 3224: It's probably better to return `ErrSTARSIllegalPosition` here (and similarly elsewhere in the function). It looks like the calling code is either checking if the error is `nil` or not, or is returning `ErrSTARSIllegalPosition` itself.

stars.go
```diff
@@ -3175,76 +3207,86 @@ func singleScope(ctx *PaneContext, facilityIdentifier string) *Controller {
...
-		if control != "" && ((controller == "C" && toCenter) || (controller == c.FacilityIdentifier && !toCenter) ||
-			(controller == ctx.world.GetController(control).FacilityIdentifier && !toCenter)) {
-			state := sp.Aircraft[callsign]
-			state.LastKnownHandoff = ctx.world.GetController(control).Scope
-			return true, control
-		} else if len(controller) > 2 {
-			controller := singleScope(ctx, string(controller[2]))
-			if controller != nil {
-				return true, controller.Callsign
+	if controller[0] == 'C' || (haveTrianglePrefix && lc == 3) {
```
mmp, line 3217: Regarding the `lc==3` test: is this implicitly baking in the knowledge that the `len` of `STARSTriangleCharacter` is 2? (Similar question about the `lc == 5 && haveTrianglePrefix` test below.)

I think it would be better to always call `strings.TrimPrefix(controller, STARSTriangleCharacter)` right after `HasPrefix(controller, STARSTriangleCharacter)` at the start of the function (and then move `lc := len(controller)` after that . Then following code can purely use `haveTrianglePrefix` to distinguish that case and can then operate on string lengths that are based on the remaining non-triangle text.

stars.go
```diff
@@ -3133,6 +3162,9 @@ func (sp *STARSPane) handoffTrack(ctx *PaneContext, callsign string, controller
 // and route.
 func calculateAirspace(ctx *PaneContext, callsign string) (string, bool) {
```
mmp, line 3163: I think it would work better to return `*AirspaceAwareness` here (with `nil` indicating "couldn't find anything"). Returning a magic string "no target" is a little error prone / obscure (e.g. it could be misspelled here or there without that being realized), and looking at the calling code, it ends up returning an error in any case when `AirspaceAwareness` isn't found. (More comments below where it's used.)

If details of why the function failed were needed, it would be better to use an `error` and to declare a global variable for the error code or codes so that calling code could use `==` tests to see what the actual error was.

stars.go
```diff
@@ -3175,76 +3207,86 @@ func singleScope(ctx *PaneContext, facilityIdentifier string) *Controller {
 
 // Give a bool if the handoff is good and the correct syntax.
 // Also decode the controller into its regular sector (N4P -> 4P)
-func (sp *STARSPane) calculateController(ctx *PaneContext, controller, callsign string) (bool, string) {
+func (sp *STARSPane) calculateController(ctx *PaneContext, controller, callsign string) (string, error) {
```
mmp, line 3210: This function is a little tricky to understand (in part because it's inherently complex how this is resolved), but I think renaming the `controller` parameter would help: it is actually.. a TCP ID? Or is there a better word to describe what the user's input corresponds to? `controller` suggests to me at least that the variable is either a controller callsign or a `Controller` object. Anyway, I think a more distinct name that makes it more clear that this is the thing the user entered that is being mapped to a control position  would make the function easier to understand.

stars.go
mmp: Actually, I think it would be better to return just a `*Controller` from this function, with `nil` for none found. Some of the calling code turns around and calls `world.GetController() `with the returned callsign, which is an unnecessary round-trip, and otherwise just shows an illegal position error if none is found. So a `nil` pointer is sufficient to convey the only failure case here.

stars.go
checkandmate1: There is one thing about this. All errors are the same because we're returning `nil`. For example, if the ACID/BCN was entered incorrectly, the error code should be `NO FLIGHT`; but, if the receiving position does not exist, the error should be `ILL POS`. I think changing it to a string and an error would be the way to go.

stars.go
checkandmate1: The 3 and 5 did have the length of the triangle character in mind. What I did before was if `haveTrianglePrefix` was true, `lc -= 1`. I could do that here, or just trim the prefix and get the length of that. One issue I see with trimming the prefix is that `haveTrianglePrefix` is used multiple times during `calculateController`, so trimming it at the beginning may lead to some problems with checking the other handoffs that could have been done.

sim.go
checkandmate1: This would be a good idea. How should this be implemented?

stars.go
```diff
@@ -5330,6 +5340,24 @@ func (sp *STARSPane) drawTracks(aircraft []*Aircraft, ctx *PaneContext, transfor
...
+		// 	if ctrl := ctx.world.GetController(ac.TrackingController); ctrl != nil && octrl != nil &&
+		// 		ctx.world.GetController(ac.TrackingController).FacilityIdentifier == octrl.FacilityIdentifier {
+		// 		trackId = ctrl.Scope
+		// 	} else {
+		// 		trackId = "*"
+		// 	}
+		// } else {
+		// 	trackId = "*"
+		// }
+
```
mmp, line 5360: (Please remove commented-out / WIP code.)

stars.go
```diff
@@ -3378,7 +3379,11 @@ func (sp *STARSPane) setLeaderLine(ctx *PaneContext, ac *Aircraft, cmd string) e
 			}
 			return nil
 		}
-	} else if len(cmd) == 2 && cmd[0] == cmd[1] { // Global leader lines
+	} else if len(cmd) == 2 { // Global leader lines
```
mmp, line 3382: I think this change (and specifically the `strings.Contains(cmd, "0")` check isn't needed since that is checked in `numpadToDirection`)

aircraft.go
```diff
@@ -22,6 +22,7 @@ type Aircraft struct {
 	FlightPlan          *FlightPlan
 	ForceQLControllers  []string
 	PointOutHistory     []string
+	DisplayFPInfo       bool
```
mmp, line 25: As I understand it, this is more or less UI-related state the World needs when drawing the coordination window. So I'd suggest instead that `World` store a `map[string]bool` for this itself, rather than putting a world-specific thing into `Aircraft`

sim.go
```diff
@@ -1768,18 +1795,34 @@ func (s *Sim) spawnAircraft() {
...
-		ac, dep, err := s.World.CreateDeparture(airport, runway, category,
-			s.LaunchConfig.DepartureChallenge, prevDep)
-		if err != nil {
-			s.lg.Errorf("CreateDeparture error: %v", err)
-		} else {
-			s.lastDeparture[airport][runway][category] = dep
-			s.lg.Infof("%s/%s/%s: launch departure", airport, runway, category)
-			s.launchAircraftNoLock(*ac)
-			s.NextDepartureSpawn[airport] = now.Add(randomWait(rateSum, false))
+		for { // There will always be a gate, because is all gates are checked this func wont get called
```
mmp, line 1798: From before I thought that this wasn't true? Wasn't there a case something like if the scenario has 0 departures to some gate then if every other gate but that one is stopped but that one isn't, then it still won't be able to launch any aircraft?

If so, would it work to just continue to call `spawnAircraft()`, let it fail if the gate it chose is stopped, and then assume it will eventually find a gate that's open (if there is one) in the next few times...?

sim.go
mmp: (This was done in 22225c237f6b4c3f1c4c15c95b27409d957cf7db, so the check here can go.)

sim.go
```diff
@@ -821,6 +821,7 @@ type Sim struct {
 	Handoffs map[string]time.Time
 	// callsign -> "to" controller
 	PointOuts map[string]map[string]PointOut
+	// ForceQL map[string]
```
mmp, line 824: (Please delete)

mmp (commented):
Overall looks good; a few small requests..

sim.go
checkandmate1: Do you mean if there are no departures made/ coded in for that one gate?

mmp: No, I mean for example, say I launch the "KEWR 4L Departure" scenario and set an ADR of 0 for KMMU and KTEB when I create the sim. Then if I halt departures from KEWR 4L I think this will go into an infinite loop because yes, there are some gates that aren't stopped, but on the other hand a departure from them will never be launched.

checkandmate1: I tried to reproduce this and looks like it was fixed in [0eb5aec](https://github.com/mmp/vice/pull/30/commits/0eb5aec6726852ee1deccc27c8499c684668ca84).

checkandmate1:
@mmp Anything that we need from this pull request (or any others)? Or should we close them?

mmp:
I think everything's been merged from this other than a few changes you'd made to the VFR (and maybe other) system list drawing code. Closing for now!

PR #32, opened 2024-03-15
Update for 7110.126B CWT and 2.5 NM Separation

catcarbon:
1. Update CWT lookup table per 7110.126B (https://www.faa.gov/regulations_policies/orders_notices/index.cfm/go/document.information/documentID/1040468)
2. Update 2.5 NM separation reduction logic


stars.go
```diff
@@ -5747,17 +5747,17 @@ func getRecatCategory(ac *Aircraft) string {
...
 }
 
-func (sp *STARSPane) checkInTrailRecatSeparation(back, front *Aircraft) {
-	actype := func(ac *Aircraft) int {
+func (sp *STARSPane) checkInTrailCwtSeparation(back, front *Aircraft) {
+	cwtClass := func(ac *Aircraft) int {
 		perf, ok := database.AircraftPerformance[ac.FlightPlan.BaseType()]
 		if !ok {
 			lg.Errorf("%s: unable to get performance model for %s", ac.Callsign, ac.FlightPlan.BaseType())
-			return 1
```
catcarbon, line 5755: Fixed: On error or not found, return the index for `NOWGT` category (Originally returning `H` category).

stars.go
```diff
@@ -5783,36 +5783,43 @@ func (sp *STARSPane) checkInTrailRecatSeparation(back, front *Aircraft) {
...
-		[10]float32{4, 3, 3, 3, 3, 3, 3, 3, 3, 10},          // Behind E
-		[10]float32{5, 5, 5, 5, 5, 4, 4, 3, 3, 10},          // Behind D
-		[10]float32{5, 5, 3.5, 3.5, 3.5, 3, 3, 3, 3, 10},    // Behind C
-		[10]float32{5, 5, 5, 5, 5, 4, 4, 3, 3, 10},          // Behind B
-		[10]float32{8, 7, 7, 7, 7, 6, 6, 4.5, 3, 10},        // Behind A
-		[10]float32{10, 10, 10, 10, 10, 10, 10, 10, 10, 10}, // Behind NOWGT (No weight: 7110.762)
-
+	// 7110.126B TBL 5-5-2
+	// 0 value means minimum radar separation
+	cwtOnApproachLookUp := [10][10]float32{ // [front][back]
```
catcarbon, line 5788: Update table per 7110.126B. 
Use `0` to represent minimum radar separation.

stars.go
```diff
@@ -5783,36 +5783,43 @@ func (sp *STARSPane) checkInTrailRecatSeparation(back, front *Aircraft) {
...
+	cwtSeparation := cwtOnApproachLookUp[cwtClass(front)][cwtClass(back)]
 
 	state := sp.Aircraft[back.Callsign]
 	vol := back.ATPAVolume()
-	if vol.Enable25nmApproach &&
-		nmdistance2ll(vol.Threshold, state.TrackPosition()) < vol.Dist25nmApproach {
-		// Reduced separation allowed starting 10nm out, if, as per 5-5-4(i):
-		// 1. leading weight class <= trailing weight class
-		// 2. super/heavy can't be leading
+	if cwtSeparation == 0 {
```
catcarbon, line 5804: Check if eligible for 2.5 NM reduction if and only if the lookup table value is `0` (minimum radar separation).
7110.126B replaces 7110.65Z 5-5-4(j) at publication. The section it replaces is now 7110.65AA 5-5-4(i).

stars.go
```diff
@@ -5783,36 +5783,43 @@ func (sp *STARSPane) checkInTrailRecatSeparation(back, front *Aircraft) {
...
+		if vol.Enable25nmApproach &&
+			nmdistance2ll(vol.Threshold, state.TrackPosition()) < vol.Dist25nmApproach {
 
-		if fclass <= bclass && fclass <= 6 {
-			mit = 2.5
+			// between aircraft established on the final approach course
+			if back.Nav.Approach.InterceptState == HoldingLocalizer &&
+				front.Nav.Approach.InterceptState == HoldingLocalizer {
+
+				// TODO: Required separation must exist prior to applying 2.5 NM separation (TBL 5-5-2)
```
catcarbon, line 5816: Checking this requires storing a state for if two aircraft already established 2.5 NM separation on final. Not worth the complexity IMO.

mmp:
This all looks good to me, thanks for the fixes!

@checkandmate1 can you take a look as well since you wrote most of this code originally?

checkandmate1:
Yeah, this all looks good. @mmp ready for a merge?

mmp:
I just realized there's one issue but I will get this merged and fix it up in a followup commit. Regarding `front.Nav.Approach.InterceptState == HoldingLocalizer`: that won't work for aircraft on RNAV approaches. (Also, in general we try not to dig into the `Nav` struct in `Aircraft` outside of nav.go/aircraft.go, though that's not consistent everywhere.)

There is an `Aircraft` `OnApproach()` method that almost works for this case, though it includes an altitude check for whether the aircraft is below the approach altitude constraints. (It's currently used for deciding if the altitude should be checked against the MVAs.) So I'll update that to optionally check the altitude and then replace these lines with calls to `OnApproach()`.

catcarbon:
@mmp I think the `OnApproach` method will return true once an aircraft passes IAF on an RNAV approach, but the condition should really be "on final approach". I will work on adding an `OnFinalApproach` method.

mmp:
Yes, that's a good point (and sounds good!)

PR #67, opened 2024-03-15
Fix EWR East Gate Handoffs in LGA scenario

catcarbon:
1. Fix `EWR 22R` departure route (right turn to 040, handoff at 6,000).
2. Add `EWR 4L` **(LGA D4/D31 only)** departure route (right turn 060, left turn 040, right turn 130, handoff at 5,000).
3. Change `EWR` departures in `LGA L31D4` scenario to use `EWR 4L` route.

### EWR 22R route
image

### EWR 4L route
image


mmp:
Thanks!

PR #113, opened 2024-03-16
Move various STARS-related config items from ScenarioGroup to STARSFacilityAdaptation

### LINKED ISSUE #34: Move more of the JSON scenario specifiers into stars_config
Things like the scope center and range, the CA volumes, scratchpad assignments, and likely other stuff should probably go in there now that it exists.

mmp:
Now that we have that (via "stars_config"), it makes more sense to have things like the radar scope default center, the inhibit CA volumes, etc., to live under it. Also updated all of the scenarios accordingly.

The data layout is now quite incompatible with the last release, so bumped the config file version and RPC version here as well.

Fixes #34.


checkandmate1:
This all looks good!

mmp:
Thanks!

PR #115, opened 2024-03-16
Delete RejectHandoff methods

mmp:
These are no longer used with the new handoff implementation...


checkandmate1:
Yeah, it's good that we got the reject handoff function out of there.

Issue #53, opened 2024-03-15
add a delay for altitude changes


mmp:
Just like with vectors, it would be good if aircraft waited a random few seconds before starting altitude changes. (And as with vectors, this is likely to be buggy and tricky to get right.)

mmp:
Ease in/out of climbs/descents is a step toward this; rather than handling it like delayed headings we might just slow walk the initial change in rate.

PR #124, opened 2024-03-17
Add altitude Delay

svalencia014:
This PR addresses issue #53 by adding a randomized delay in a similar fashion as the heading delay system.

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -30,6 +30,10 @@ type Nav struct {
 	// followed, it's fine for the second to override it.
 	DeferredHeading *DeferredHeading
 
+	// Deferred Altitude does the same as DeferredHeading, but for altitude
+	// assignements
+	DeferredAltitude *DeferredAltitude
+
 	FinalAltitude float32
 	Waypoints     []Waypoint
 }
@@ -47,6 +51,12 @@ type DeferredHeading struct {
 	// intended.
 	Time    time.Time
 	Heading NavHeading
+}
+
+type DeferredAltitude struct {
+	// Same as DeferredHeading, but for altitude assignments
+	Time     time.Time
+	Altitude NavAltitude
 }
 
 type FlightState struct {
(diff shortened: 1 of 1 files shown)


mmp:
More places to consider:

- `ExpediteDescent()` and `ExpediteClimb()`: they shouldn't issue that 'unable' message if there is a pending altitude change. (Maybe in that case they should start the altitude change immediately anyway.)
- `CrossFixAt()` where it does `nav.Altitude = NavAltitude{}`
- Halfway through `clearedApproach()` in the `nav.Approach.InterceptState == HoldingLocalizer` test.
- ClimbViaSID, DescendViaSTAR, and a bunch of places in the procedure turn code.

More generally, the tricky thing about this (as was the case with getting deferred turns to work correctly) is that it's necessary to audit all of the code that looks at `nav.Altitude`, either setting it or clearing it, and think through the implications of having a `DeferredAltitude` that's going to clobber it a few seconds in the future. Sometimes the action should be based on the deferred altitude, sometimes it's something that should override/clear out the deferred altitude, etc...

svalencia014:
For the first point, the delay would still make sense based on reaction time of the pilots (A pilot wouldn't instantly start dropping altitude as soon as the instruction is transmitted, they would still need to dial it in).

mmp:
(Just to clarify, are you saying that changes to the code make sense in those cases, or are you saying you think they're all ok as currently implemented?)

svalencia014:
I would say that leaving those other implementations as is would be fine. I had a few people play-test a build I made with the altitude delay before opening this PR and they noticed no problems.

mmp:
With your branch, if I give an aircraft a descent and then right afterward give them `ED` to expedite the descent, I get:
image
So they read back the descent and then claim they aren't descending. In the real world, even if they're still dialing in the new altitude, they're not going to say they're not descending.

So I think that general issue applies to many places in the code: there is a check "is the aircraft changing altitude; if so do X, else do Y" and the check will think they are not changing altitude even though they have been given an instruction to do so...

PR #127, opened 2024-03-17
(Re-)allow specifying multiple sector ids for quicklook

### LINKED ISSUE #125: multiple sector QL doesn't work in the development version
`[multifunc]1L 1V[enter]` should quick look both sector ids but currently gives `ILL POS`

mmp:
Fixes #125

Note that the old `parseQuickLookPositions()` had some functionality that I think is now taken care of by `calculateController()`, but please take a look and confirm. I also factored the logic into a stand-alone function to reduce code duplication a bit.


checkandmate1:
This all looks good. Thanks Matt

PR #131, opened 2024-03-17
130 lib scenarios have incorrect altitude for some flight plans

checkandmate1:
Fixed the LIb flight plans and then added some checking for low altitudes. @mmp let me know if it's ready for a merge

### DIFF
--- a/sim.go
+++ b/sim.go
@@ -1743,6 +1743,9 @@ func (s *Sim) spawnAircraft() {
 			if ac, err := s.World.CreateArrival(group, arrivalAirport, goAround); err != nil {
 				s.lg.Error("CreateArrival error: %v", err)
 			} else if ac != nil {
+				if ac.FlightPlan.Altitude < 500 { // Just so altitudes of 500 can be filled. (For VFR?)
+					ac.FlightPlan.Altitude *= 100
+				}
 				s.launchAircraftNoLock(*ac)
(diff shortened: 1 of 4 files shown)


sim.go
```diff
@@ -1743,6 +1743,9 @@ func (s *Sim) spawnAircraft() {
 			if ac, err := s.World.CreateArrival(group, arrivalAirport, goAround); err != nil {
 				s.lg.Error("CreateArrival error: %v", err)
 			} else if ac != nil {
+				if ac.FlightPlan.Altitude < 500 { // Just so altitudes of 500 can be filled. (For VFR?)
```
mmp, line 1746: Rather than patching it up here I'd suggest detecting bogus altitudes when vice starts up and issuing an error then; that way if this happens in the future, people developing scenarios can know to fix it themselves.

So for departures, if you go to `Airport` `PostDeserialize()` in airport.go, around line 433 it loops over the departures and does various validations. You could check `dep.Altitude` there and use the `ErrorLogger` to report any issues.

website/index.html
```diff
@@ -1,35 +1,39 @@
 <!DOCTYPE html>
-<html lang="en"> 
+<html lang="en">
```
mmp, line 2: Is it intended to include index.html changes in this PR?

PR #137, opened 2024-03-18
EWR Exit Categories wrong in the LGA file.

checkandmate1:
MERIT & GREKI were never added in the file, making them never spawn. They were added


website/index.html
```diff
@@ -1,35 +1,39 @@
 <!DOCTYPE html>
-<html lang="en"> 
+<html lang="en">
```
catcarbon, line 2: What happened with the formatter? It seems to be changing blank spaces every time index.html is modified...

mmp: More generally, are there any intended changes to index.html here? If not, please remove it from the PR. If so, they should go in without reformatting the whole HTML file.

PR #138, opened 2024-03-18
Fixed #135: 2.5 NM Separation Reduction and ATPA Alert/Warning Fix

### LINKED ISSUE #135: OnFinalApproach method for 2.5 NM separation
1. Add `OnFinalApproach` method for 2.5 NM separation processing (See #32).
2. ATPA warning/alert isn't working correctly: all 45 iterations of separation calculation produce the same value (see image below).

Screenshot 2024-03-18 at 11 00 54 PM

catcarbon:
Link to issue: #135

### DIFF
--- a/nav.go
+++ b/nav.go
@@ -307,6 +307,10 @@ func (nav *Nav) OnApproach(checkAltitude bool) bool {
 	return false
 }
 
+func (nav *Nav) OnFinalApproach() bool {
+	return false
+}
+
 ///////////////////////////////////////////////////////////////////////////
 // Communication
 
--- a/nav.go
+++ b/nav.go
@@ -308,7 +308,12 @@ func (nav *Nav) OnApproach(checkAltitude bool) bool {
 }
 
 func (nav *Nav) OnFinalApproach() bool {
-	return false
+	if !nav.OnApproach(false) {
+		return false
+	}
+
+	_, hasPassedFaf, err := nav.distanceToEndOfApproach()
+	return err == nil && hasPassedFaf
 }
 
 ///////////////////////////////////////////////////////////////////////////
@@ -1224,7 +1229,7 @@ func (nav *Nav) getWaypointAltitudeConstraint() *WaypointCrossingConstraint {
 func (nav *Nav) TargetSpeed(lg *Logger) (float32, float32) {
 	maxAccel := nav.Perf.Rate.Accelerate * 30 // per minute
 
-	fd, err := nav.finalApproachDistance()
+	fd, _, err := nav.distanceToEndOfApproach()
 	if err == nil && fd < 5 {
 		// Cancel speed restrictions inside 5 mile final
 		lg.Debug("speed: cancel speed restrictions at 5 mile final")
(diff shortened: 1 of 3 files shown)


catcarbon:
Link to issue: #135

catcarbon:
Questions to be answered:
1. Are there conditions that revert 2.5 NM separation to 3 NM? (e.g. leading aircraft goes around)

catcarbon:
ATPA warning/alert fix:

### Warning state
Screenshot 2024-03-18 at 11 32 37 PM

### Alert state
Screenshot 2024-03-18 at 11 32 53 PM

### Debug log
Screenshot 2024-03-18 at 11 33 07 PM

stars.go
```diff
@@ -5859,16 +5859,17 @@ func (sp *STARSPane) checkInTrailCwtSeparation(back, front *Aircraft) {
...
-	bac := MakeModeledAircraft(back, state, vol.Threshold)
+	frontModel := MakeModeledAircraft(front, sp.Aircraft[front.Callsign], vol.Threshold)
+	backModel := MakeModeledAircraft(back, state, vol.Threshold)
 
 	// Will there be a MIT violation s seconds in the future?  (Note that
 	// we don't include altitude separation here since what we need is
 	// distance separation by the threshold...)
-	fp, bp := fac.p, bac.p
-	for s := float32(0); s < 45; s++ {
-		fp, bp := fac.NextPosition(fp), bac.NextPosition(bp)
```
catcarbon, line 5870: Short assignment `:=` accidentally shadowed the variables outside of for-loop.

nav.go
```diff
@@ -1357,33 +1366,43 @@ func (nav *Nav) getUpcomingSpeedRestrictionWaypoint() (*Waypoint, float32, float
...
-		d := nmdistance2ll(nav.FlightState.Position, wp[0].Location)
+		remainingDistance := nmdistance2ll(nav.FlightState.Position, wp[0].Location)
 		for i := 0; i < len(wp)-1; i++ {
-			d += nmdistance2ll(wp[i].Location, wp[i+1].Location)
+			remainingDistance += nmdistance2ll(wp[i].Location, wp[i+1].Location)
+		}
+		// Check if any remaining waypoint is the FAF, in reverse direction.
+		isPastFaf := true
+		for i := len(wp) - 1; i > -1; i-- {
+			if wp[i].FAF {
```
mmp, line 1400: We don't currently require that `/faf` be specified with approaches; it previously had only been used for labeling the fix when drawing an approach on the STARs scope. 

There's a commented-out check in `WaypointArray` `CheckApproach()` that prints an error if this is missing; it seems that a fair number of approaches are missing them, though a fair number also have them.

Fortunately, only a few N90 scenarios currently have 2.5nm on final enabled for ATPA. So I would suggest removing that commented-out code from `CheckApproach()` and then adding an equivalent check inside the `for rwy, vol := range ap.ATPAVolumes {` loop in `Airport` `PostDeserialize()`, which is part of the sanity checks done during loading. That way as facility engineers enable this at other places, they'll be notified of the need to add the `/faf` specifier.

(That seems preferable to requiring it everywhere and then needing to fix all of the scenarios now.)

Related, it looks like the ARINC424 parser isn't getting the FAF; I filed a [tracking bug](https://github.com/mmp/vice/issues/140).

stars.go
```diff
@@ -613,7 +613,7 @@ type STARSPreferenceSet struct {
 	AutomaticFDBOffset       bool
 
 	DisplayTPASize               bool
-	DisplayATPAIntrailDist       bool
+	DisplayATPAInTrailDist       bool
```
mmp, line 616: I'm all for renaming this but a note that doing so will cause the setting to be lost in users' saved configs. (Basically vice takes advantage of `encoding/json`'s ability to have a big nested struct thrown at it and serialize it all out in order to save/restore that stuff.) So if a user had set the old version of this to `true`, then upon deserialization that would be discarded since there's no DisplayATPAIn**t**railDist variable any more and DisplayATPAIn**T**railDist would be default-initialized to false.

You might add: `` `json:"DisplayATPAIntrailDist"` `` here so that the old spelling is still used for saves so that this still works.

mmp (commented):
Looks great--two small things.

nav.go
catcarbon: BN clarified for me that whether an aircraft passed FAF or not on approach is irrelevant to the "on final approach course" requirement. So I will go ahead and remove this check.

mmp: Ah, that's good to know and nice that it makes things easier here.

mmp:
Looks great!

PR #154, opened 2024-03-22
Warnings reorg

mmp:


checkandmate1 (changes requested):
We're still encountering SPC's overriding each other with a new one. For example, if there is an RF on an aircraft, and an EM is added, that EM Will override the previous RF,

stars.go
```diff
@@ -5945,51 +5946,57 @@ func (sp *STARSPane) diverging(a, b *Aircraft) bool {
 	return true
 }
 
-func (sp *STARSPane) formatDatablocks(ctx *PaneContext, ac *Aircraft) []STARSDatablock {
+func (sp *STARSPane) getWarnings(ctx *PaneContext, ac *Aircraft) []string {
+	var warnings []string
```
checkandmate1, line 5950: When `warnings` is created it never accounts for the warnings that are not in the string `SPCOveride`. What ends up happening is that only the warning that is in SPCOverride is shown as a warning. Either converting SPCOverride to a []string, so that it can store multiple warnings, or replacing `warnings` with `state.warnings` so that the previous warnings can save. Then just having some logic to prevent duplicate entries:
```go
if index := slices.Index(state.Warnings, state.SPCOverride); index != -1 {
		state.Warnings = append(state.Warnings[:index], state.Warnings[index+1:]...)
	} else if ac.Squawk == Squawk(0o7700) || state.SPCOverride == "EM" {
state.Warnings = append(state.Warnings, "EM")
} // etc...
```

stars.go
```diff
@@ -5945,51 +5946,57 @@ func (sp *STARSPane) diverging(a, b *Aircraft) bool {
 	return true
 }
 
-func (sp *STARSPane) formatDatablocks(ctx *PaneContext, ac *Aircraft) []STARSDatablock {
+func (sp *STARSPane) getWarnings(ctx *PaneContext, ac *Aircraft) []string {
```
checkandmate1, line 5949: I do think `getWarnings` is a good thing to have, just to clean `formatDatablocks` up a bit

stars.go
```diff
@@ -5945,51 +5946,57 @@ func (sp *STARSPane) diverging(a, b *Aircraft) bool {
...
+		warnings = append(warnings, "HJ")
+	}
+	if ac.Squawk == Squawk(0o7600) || state.SPCOverride == "RF" {
+		warnings = append(warnings, "RF")
+	}
+	if ac.Squawk == Squawk(0o7700) || state.SPCOverride == "EM" {
+		warnings = append(warnings, "EM")
+	}
+	if ac.Squawk == Squawk(0o7777) || state.SPCOverride == "MI" {
+		warnings = append(warnings, "MI")
```
checkandmate1, line 5967: I think MI should be replaced with LL (lost link). I got told that MI was also yellow, but I'm not 100% sure.

stars.go
mmp: Ah, got it. I changed to a `map[string]interface{}` to be able maintain multiple overrides and also added logic so that specifying an active override toggles it back off.

stars.go
mmp: p. 2-66 of the STARS manual has 7400 for LL and 7777 for MI.

The change I pushed reworks handling of SPCs significantly both making it simpler and more extensible (I think). I left off making MI yellow since that will take some further work to the datablock formatting code. But I think the `SPC` struct there could just be extended to hold a color for each one and then we could access that in the datablock code in the future...

stars.go
checkandmate1: Changing it to a map may cause some formatting issues, AFAIK maps aren't ordered like slices/ arrays.

mmp: Indeed, maps give the keys in random order, but everywhere it's used it goes through `SortedMapKeys` so they will come out in a consistent order.

checkandmate1 (approved):
This all looks good! Using a map[string]interface{} is definitely the way to go (also naturally protects against duplicate entries)

PR #156, opened 2024-03-22
STARSPane: rework global leader line handling and polish leader line details

mmp:
For global leader lines, we need to be aware of when they are set since (AFAICT) if a global leader line direction is set for an aircraft, it overrides a controller's local per-aircraft leader line direction and similarly, if a controller sets a per-ac leader line direction, it overrides a global leader line direction.

Therefore:
- Setting the global leader line direction for an aircraft now posts an event to EventStream; STARSPane pays attention to this.
- STARSAircraftState now stores a bool UseGlobalLeaderLine to indicate whether to use the local or global one if both are set.

Other fixes:
- [Multifunc]L5* clears the 'controlled by other controllers' leader line direction setting (4-98)
- Leader line commands now use calculateController to look up TCPs (and not World GetController!)
- Cleaned up the code that distinguishes between setting a per-aircraft leader line direction specified using ACID ([Multifunc]L(#) (ACID)) and a per-controller leader line specified via TCP ([Multifunc]L(TCP)(#)).

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -408,11 +408,13 @@ type STARSAircraftState struct {
 	LastKnownHandoff         string
 	POFlashingEndTime        time.Time
 
-	// This is only set if a leader line direction was specified for this
-	// aircraft individually
-	LeaderLineDirection *CardinalOrdinalDirection
-	GlobalLeaderLine    bool // If the aircraft is a global line or set by the TCP
-	Ghost               struct {
+	// These are only set if a leader line direction was specified for this
+	// aircraft individually:
+	LeaderLineDirection       *CardinalOrdinalDirection
+	GlobalLeaderLineDirection *CardinalOrdinalDirection
+	UseGlobalLeaderLine       bool
+
+	Ghost struct {
 		PartialDatablock bool
 		State            GhostState
 	}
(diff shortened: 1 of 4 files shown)


checkandmate1:
This all looks good.

stars.go
```diff
@@ -2318,75 +2330,68 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
 		case "L":
 			// leader lines
-			if l := len(cmd); l == 0 {
-				status.err = ErrSTARSCommandFormat
-				return
-			} else if l == 1 {
+			if l := len(cmd); l == 1 {
 				if dir, ok := numpadToDirection(cmd[0]); ok && dir != nil {
-					// Tracked by me
+					// 4-97: tracked by me, '5' not allowed
```
checkandmate1, line 2335: When entering 5, no error is returned. 4-97 says that `FORMAT` should be returned. Same thing with line 2343

mmp: I think this is actually ok (though it's a little obscure): `numpadToDirection()` return (nil, true) if '5' is entered; so this (and 2343) should fall into the `FORMAT` case.

Issue #155, opened 2024-03-22
inter-facility handoff not working


mmp:
If I start a KEWR scenario and enter `Δ1A` and slew to hand off to ABE_APP, nothing happens: the command is accepted without an error but the handoff never happens, nothing shows up in the datablock to indicate an outbound handoff, etc.

checkandmate1:
∆1A shouldn't work from N90 to ABE. The correct syntax would be ∆4 or ∆41A/D. 1 is the facility identifier for PVD AFAIK; however, that's weird about how it does nothing. Was it inputted to the scratchpad?

checkandmate1:
I am seeing it now. It does nothing, it should input that into the scratchpad.

PR #157, opened 2024-03-22
Fixed weird scratchpads

checkandmate1:
Just addressing #155. I fixed the scratchpad logic a bit, just so some weird ones could be entered. (Like ∆1A)

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -3045,7 +3045,15 @@ func (sp *STARSPane) updateQL(ctx *PaneContext, input string) (ok bool, previewI
 }
 
 func (sp *STARSPane) setScratchpad(ctx *PaneContext, callsign string, contents string, isSecondary bool) error {
-	if len(contents) > 4 {
+	lc := len(contents)
+	if strings.Contains(contents, STARSTriangleCharacter) {
+		lc -= 1
+	}
+	var index int 
+	if isSecondary {
+		index = 1
+	}
+	if lc > 4 || (lc > 3 && !ctx.world.STARSFacilityAdaptation.ScratchpadRules[index]) {
 		return ErrSTARSIllegalScratchpad
 	}
 
@@ -3319,16 +3327,6 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 	ac, acDistance := sp.tryGetClosestAircraft(ctx.world, mousePosition, transforms)
 	ghost, ghostDistance := sp.tryGetClosestGhost(ghosts, mousePosition, transforms)
 
-	isControllerId := func(id string) bool {
-		// FIXME: check--this is likely to be pretty slow, relatively
-		// speaking...
-		for _, ctrl := range ctx.world.GetAllControllers() {
-			if ctrl.SectorId == id {
-				return true
-			}
-		}
-		return false
-	}
 
 	ps := &sp.CurrentPreferenceSet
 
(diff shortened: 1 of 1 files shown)


stars.go
```diff
@@ -3045,7 +3045,45 @@ func (sp *STARSPane) updateQL(ctx *PaneContext, input string) (ok bool, previewI
...
+		}
+		if !found {
+			return ErrSTARSCommandFormat
+		}
+	}
+
+	illegalScratchpads := []string{"NAT", "CST", "AMB", "RDR", "ADB", "XXX"}
+	if slices.Contains(illegalScratchpads, contents) {
 		return ErrSTARSIllegalScratchpad
 	}
```
checkandmate1, line 3088: This is all error-checking for the scratchpads. The rules for the scratchpads can be found on 5-156

stars.go
```diff
@@ -3732,17 +3742,13 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 					}
 					// Otherwise fall through to try running it as a command
 				}
-
 				// If it didn't match a controller, try to run it as a command
 				ctx.world.RunAircraftCommands(ac, cmd,
 					func(err error) {
 						// If it's not a valid command and fits the requirements for a scratchpad, set the scratchpad.
-						if len(cmd) <= 3 || (len(cmd) >= 4 && ctx.world.STARSFacilityAdaptation.ScratchpadRules[0]) {
-							if err := sp.setScratchpad(ctx, ac.Callsign, cmd, false); err != nil {
```
checkandmate1, line 3741: All of this isn't needed anymore because of the error checking done in `STARSPane setScratchpad`

stars.go
```diff
@@ -3045,7 +3045,45 @@ func (sp *STARSPane) updateQL(ctx *PaneContext, input string) (ok bool, previewI
 }
 
 func (sp *STARSPane) setScratchpad(ctx *PaneContext, callsign string, contents string, isSecondary bool) error {
-	if len(contents) > 4 {
+	lc := len(contents)
+	if strings.Contains(contents, STARSTriangleCharacter) {
+		lc -= 1
+	}
+	if ac := ctx.world.GetAircraft(callsign, false); ac != nil && ac.TrackingController == "" {
+		return ErrSTARSIllegalTrack /* This is because /OK can be used for associated tracks that are not owned by this TCP. But /OK cannot be used
```
checkandmate1, line 3053: The `/OK` logic will have to be in `Sim SetScratchpad` because it calls `DispatchTrackingCommand`, so well need some logic there to bypass that

stars.go
```diff
@@ -3045,7 +3045,45 @@ func (sp *STARSPane) updateQL(ctx *PaneContext, input string) (ok bool, previewI
...
+		"A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M",
+		"N", "O", "P", "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z",
+		"0", "1", "2", "3", "4", "5", "6", "7", "8", "9", STARSTriangleCharacter,
+		".", "/", "*",
+	}
+
+	for _, letter := range contents {
+		var found bool
+		for _, allowed := range allowedCharacters {
+			if strings.Contains(allowed, string(letter)) {
```
mmp, line 3075: I think this ends up checking whether *any* character is valid, while I think you want to check that all are valid. What about:
```
allowedCharacters := "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789./*" + STARSTriangleCharacter
for _, letter := range contents {
    if !strings.ContainsRune(allowedCharacters, letter) {
        return ErrSTARSCommandFormat
    }
}
```

stars.go
```diff
@@ -3045,7 +3045,45 @@ func (sp *STARSPane) updateQL(ctx *PaneContext, input string) (ok bool, previewI
 }
 
 func (sp *STARSPane) setScratchpad(ctx *PaneContext, callsign string, contents string, isSecondary bool) error {
-	if len(contents) > 4 {
+	lc := len(contents)
+	if strings.Contains(contents, STARSTriangleCharacter) {
```
mmp, line 3049: Instead how about just `lc := len([]rune(contents))`? It's slightly obscure to implicitly be using the knowledge that the triangle is two bytes in the string and hardcoding just triangle here wouldn't cover the case of any potential future non-single-byte characters (not that that seems likely, though...)

mmp (commented):
Sorry for not getting to this sooner. Looks good overall. Two small things.

PR #159, opened 2024-03-24
Fix handling of various CRDA commands

### LINKED ISSUE #39: CRDA command syntax bug
Via @checkandmate1 

```
> The Toggle display of a runway's CRDA qualification region syntax is
> supposed to be <MULTI FUNC> <N> [Optional <AIRPORT IDENTIFIER> <SPACE>]
> <RUNWAY NUMBER> <SPACE> <B> <ENTER> Currently, there is no space between
> the airport identifier and the runway number.
```

mmp:
- Correctly parse optional airports (fixes #39).
- Use controller default airport when airport is unspecified (for commands that use it.)
- Add [MF]NP(airport)(#) to toggle ghosts for a runway pair
- Fix CRDA slew commands to match the specification
- Added a few STARS error codes used for CRDA
- Updated docs

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -109,10 +109,13 @@ var (
 	ErrSTARSIllegalCode       = NewSTARSError("ILL CODE")
 	ErrSTARSIllegalFix        = NewSTARSError("ILL FIX")
 	ErrSTARSIllegalFlight     = NewSTARSError("ILL FLIGHT")
+	ErrSTARSIllegalFunction   = NewSTARSError("ILL FNCT")
 	ErrSTARSIllegalLine       = NewSTARSError("ILL LINE")
 	ErrSTARSIllegalMap        = NewSTARSError("ILL MAP")
 	ErrSTARSIllegalParam      = NewSTARSError("ILL PARAM")
 	ErrSTARSIllegalPosition   = NewSTARSError("ILL POS")
+	ErrSTARSIllegalRPC        = NewSTARSError("ILL RPC") // CRDA runway pair config
+	ErrSTARSIllegalRunway     = NewSTARSError("ILL RWY")
 	ErrSTARSIllegalScratchpad = NewSTARSError("ILL SCR")
 	ErrSTARSIllegalSector     = NewSTARSError("ILL SECTOR")
 	ErrSTARSIllegalText       = NewSTARSError("ILL TEXT")
--- a/stars.go
+++ b/stars.go
@@ -2395,6 +2395,11 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
 
 		case "N":
 			// CRDA...
+			if len(sp.ConvergingRunways) == 0 {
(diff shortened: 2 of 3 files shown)


errors.go
```diff
@@ -109,10 +109,13 @@ var (
 	ErrSTARSIllegalCode       = NewSTARSError("ILL CODE")
 	ErrSTARSIllegalFix        = NewSTARSError("ILL FIX")
 	ErrSTARSIllegalFlight     = NewSTARSError("ILL FLIGHT")
+	ErrSTARSIllegalFunction   = NewSTARSError("ILL FNCT")
```
checkandmate1, line 112: This should be here, but also `ILL FUNC` is also a thing as well. (5-161). Not exactly sure what the difference is.

stars.go
```diff
@@ -2433,133 +2430,210 @@ func (sp *STARSPane) executeSTARSCommand(cmd string, ctx *PaneContext) (status S
...
+				} else if cmd[0] == 'P' {
+					// These commands either start with an airport and a
+					// space or use the controller's default airport if
+					// none is specified. None of the commands otherwise
+					// allow spaces, so we can use the presence of a space
+					// to determine if an airport was specified.
+					airport, extra, ok := strings.Cut(cmd[1:], " ")
+					if !ok {
+						ctrl := ctx.world.GetController(ctx.world.Callsign)
+						airport = ctrl.DefaultAirport[1:] // drop leading "K"
```
checkandmate1, line 2562: I already managed to break it. Entering the command `MULTI FUNC NPEWR1` seemed to crash it because the slice [1:0] was out of bounds. I'm not 100% sure how this command works, but it is good that I found this bug out

checkandmate1:
Where in the STARS manual is the NP command?

checkandmate1 (changes requested):
Using the command `MULTI FUNC NEWR1T` gave an `ILL PARAM`, so that may be something to look into. Also, I can't get the NP command to work either.

mmp:
> Where in the STARS manual is the NP command?

p6-22, at least in the copy I have. ("Toggle display of Ghost data blocks for specified runway pair")

mmp:
> Using the command `MULTI FUNC NEWR1T` gave an `ILL PARAM`, so that may be something to look into. Also, I can't get the NP command to work either.

Assuming you mean to enable tie mode, I believe it should be `[MULTIFUNC]NPEWR 1T` (8-11) otherwise it's `[MULTIFUNC]NEWR 1E` to enable ghosts (6-23). So I think that's a valid error.

stars.go
mmp: I'm not able to reproduce this. I'm wondering if you tried with a saved session rather than starting a new one. (If the saved session was from before `Controller` `DefaultAirport` was introduced then it would be unset and this would happen. In the latest version we validate that when scenarios are loaded when vice launches: https://github.com/mmp/vice/blob/463be18bacddf1a9dea6e1c7fbecd02d011561f7/scenario.go#L298.

Issue #168, opened 2024-04-01
Fix live weather


checkandmate1:

mmp:
Not sure if it's overkill but using something like https://github.com/urkk/metar for the METAR parsing could be worthwhile.

checkandmate1:
What only parses a METAR, it doesn't get you the METAR from an ICAO code correct?

mmp:
Yes, I believe so.

checkandmate1:
The only METAR parsing that is going on is finding the altimeter. (For some reason it's in HPA and not INHG). I don't think there is a reason to parse the already parsed METAR. If I'm overlooking something please let me know!

mmp:
Ah, got it, sorry. I was thinking the crash was from parsing the wind from the METAR. So nevermind..

PR #169, opened 2024-04-01
Fix live weather

checkandmate1:
Hopefully should be a fix to #168


mmp:
Looking again at the AirportWeatherData code, should this be happening for all entries in data? i.e. `for i := range data { if data[i].Wdir == "VRB" { ...` ?

checkandmate1:
The length is always one, so I believe it's safe to do that.

Issue #171, opened 2024-04-02
Fix scratchpads


checkandmate1:

mmp:
@checkandmate1 I think this can be closed?

checkandmate1:
Yes.

PR #172, opened 2024-04-02
Fixed long scratchpads & added error checking for too long scratchpads

checkandmate1:
Fixing #171 

- _vice_ will now shoot an error for invalid departure scratchpads.
- Fixed the departure scratchpads for some scenarios


airport.go
```diff
@@ -582,6 +582,16 @@ func (ap *Airport) PostDeserialize(icao string, sg *ScenarioGroup, e *ErrorLogge
 			}
 		}
 	}
+	for _, scratchpad := range sg.STARSFacilityAdaptation.Scratchpads {
```
mmp, line 585: Should this be `for i, scratchpad := ...` and then `ScratchpadRules[i]` below (2x) ?

checkandmate1: `i` would be a string. `sg.STARSFacilityAdaptation.Scratchpads` is a `map[string][string]`. All departure scratchpad (AFAIK) are primary, so I can assume that it's going to be the primary one.

checkandmate1:
@mmp Should this PR be closed, or is anything stopping me from merging it in for the release? (Or did we add in these features already in other commits?)

mmp:
I think this got merged (and then I added a few more rules that I noticed in the STARS manual): https://github.com/mmp/vice/blob/d3fdaa1c09cfb77dd5463252b8122456b283cd55/stars.go#L3115. Though please double-check that nothing was missed..

checkandmate1:
For the part checking implied commands and seeing if it matches a TCP, I think it should first check the facility. 
```
if ctrl.FaciliyIdentifier == "" && ctrl.SectorId == contents {
// etc...
}
```

mmp:
> For the part checking implied commands and seeing if it matches a TCP, I think it should first check the facility.
> 
> ```
> if ctrl.FaciliyIdentifier == "" && ctrl.SectorId == contents {
> // etc...
> }
> ```

Yep, good catch. Done and pushed. I will go ahead and close this but reopen if I've missed something.

PR #177, opened 2024-04-11
add video maps to D21

golfyplayz:
Add more video maps to D21 scenario, such as MVA, airspace delegation, and sids. It may say that some changes were made to M98, but there is nothing changed, I just can't get it the same where Github doesn't say I changed anything


checkandmate1:
Thanks!

Issue #75, opened 2024-03-16
Fix history trails


checkandmate1:
History trail pulls the histories instead of adding a new one and simultaneously removing one. Reference: [History Trails](https://discord.com/channels/1090685318014976081/1218380387001040946/1218380387001040946)

PR #183, opened 2024-04-27
STARSPane: fix bugs in handling of track history

mmp:
Now we do a better job of distinguishing between the last track from the radar (which is now stored in STARSAircraftState track) and the track history, which is stored in STARSAircraftState historyTracks. We follow STARS better in that there is now a history update rate (set via the H_RATE DCB button) and the history tracks are updated the first time we get a new radar track and the history rate number of seconds have passed.

Making this distinction for example lets us use the most recent radar track altitude in datablocks without regard to the track history (and its update rate.)

Also fixed history drawing to not draw history tracks at all if the associated brightness is zero.

Fixes issue #75.

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -127,8 +127,9 @@ type STARSPane struct {
 
 	HavePlayedSPCAlertSound map[string]interface{}
 
-	lastTrackUpdate time.Time
-	discardTracks   bool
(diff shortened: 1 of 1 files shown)


checkandmate1 (approved):
Looks and works well.

PR #187, opened 2024-04-28
Global Chat

checkandmate1:
Global chat can be used with the `/` character at the start of the message. Auto capitalization is removed when the first character of the input is a slash.

### DIFF
--- a/eventstream.go
+++ b/eventstream.go
@@ -191,6 +191,7 @@ const (
 	RadioTransmissionEvent
 	StatusMessageEvent
 	ServerBroadcastMessageEvent
+	GlobalMessageEvent
 	AcknowledgedPointOutEvent
 	RejectedPointOutEvent
 	IdentEvent
--- a/eventstream.go
+++ b/eventstream.go
@@ -204,7 +204,7 @@ const (
 func (t EventType) String() string {
 	return []string{"InitiatedTrack", "DroppedTrack", "PushedFlightStrip", "PointOut",
 		"OfferedHandoff", "AcceptedHandoff", "CanceledHandoff", "RejectedHandoff",
-		"RadioTransmission", "StatusMessage", "ServerBroadcastMessage",
+		"RadioTransmission", "StatusMessage", "ServerBroadcastMessage", "GlobalMessage",
 		"AcknowledgedPointOut", "RejectedPointOut", "Ident", "HandoffControll",
 		"SetGlobalLeaderLine", "TrackClicked"}[t]
 }
--- a/panes.go
+++ b/panes.go
@@ -1009,6 +1009,7 @@ func (mp *MessagesPane) runCommands(w *World) {
 	if mp.input.cmd[0] == '/' {
 		w.SendGlobalMessage(w.Callsign + ": " + mp.input.cmd[1:])
 		mp.messages = append(mp.messages, Message{contents: w.Callsign + ": " + mp.input.cmd[1:], global: true,})
+		mp.history = append(mp.history, mp.input)
 		mp.input = CLIInput{}
 		return
 	}
--- a/panes.go
+++ b/panes.go
@@ -787,6 +787,7 @@ type Message struct {
 	contents string
 	system   bool
 	error    bool
+	global 	 bool
 }
 
 type CLIInput struct {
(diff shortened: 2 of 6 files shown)


eventstream.go
```diff
@@ -191,6 +191,7 @@ const (
 	RadioTransmissionEvent
 	StatusMessageEvent
 	ServerBroadcastMessageEvent
+	GlobalMessageEvent
```
mmp, line 194: The `String()` method below needs to be updated to account for this.

world.go
```diff
@@ -30,9 +30,13 @@ type World struct {
 	METAR       map[string]*METAR
 	Controllers map[string]*Controller
 
+	GlobalMessage GlobalMessage
+
 	DepartureAirports map[string]*Airport
 	ArrivalAirports   map[string]*Airport
 
+	ERAMComputers map[string]ERAMFacilityData
```
mmp, line 38: Not intended for this PR, I assume.

world.go
```diff
@@ -89,6 +93,15 @@ type World struct {
 	STARSInputOverride string
 }
 
+type ERAMFacilityData struct {
```
mmp, line 96: These also should be moved out of the PR

sim.go
```diff
@@ -1276,10 +1276,16 @@ func (s *Sim) PostEvent(e Event) {
...
+}
+
 type SimWorldUpdate struct {
-	Aircraft    map[string]*Aircraft
-	Controllers map[string]*Controller
-	Time        time.Time
+	Aircraft      map[string]*Aircraft
+	Controllers   map[string]*Controller
+	Time          time.Time
+	GlobalMessage GlobalMessage
```
mmp, line 1288: I don't think `GlobalMessage` here is used since it's put out via the event stream?

world.go
```diff
@@ -30,9 +30,13 @@ type World struct {
 	METAR       map[string]*METAR
 	Controllers map[string]*Controller
 
+	GlobalMessage GlobalMessage
```
mmp, line 33: Likewise I think this is unused?

panes.go
```diff
@@ -997,6 +1005,17 @@ func (mp *MessagesPane) processKeyboard(ctx *PaneContext) {
 func (mp *MessagesPane) runCommands(w *World) {
 	callsign, cmd, ok := strings.Cut(mp.input.cmd, " ")
```
mmp, line 1006: Since this isn't used in the 'send message' case it would be good to move it below the message sending `if` test below so that it's closer to where those variables are actually used.

panes.go
```diff
@@ -902,6 +904,8 @@ func (mp *MessagesPane) Draw(ctx *PaneContext, cb *CommandBuffer) {
 		msg := mp.messages[len(mp.messages)-1-i]
 		s := Select(msg.error, errorStyle, Select(msg.system, systemStyle, style))
 		td.AddText(msg.contents, [2]float32{indent, y}, s)
+		g := Select(msg.global, globalStyle, Select(msg.system, systemStyle, style))
+		td.AddText(msg.contents, [2]float32{indent, y}, g)
```
mmp, line 908: We've got two `AddText` calls now where the second will draw over the first one. In turn I think error messages will be redrawn with systemStyle here.

Since the style selection is getting a little complicated, what about something like adding a `Color()` method to `Message` that returns an `RGB` where it returns a color based on whether `system` is set, `error` is set, etc. Then this just does something like:
```
s := TextStyle{Font: mp.Font, Color: msg.Color()}
td.AddText(..., s)
```

mmp (commented):
Looks good overall. Can you also update the docs in `website/index.html` to mention this in an appropriate place and also increment `ViceRPCVersion` in server.go since a new RPC call has been added so we don't want people with old clients connecting to the server and wondering why they're not getting messages.

website/index.html
```diff
@@ -1736,13 +1736,9 @@ <h3>Handing off Aircraft</h3>
...
             <p>When you are ready to transfer control of the aircraft to the controller who has
-              accepted the track, slew the aircraft's radar track. The datablock will turn green,
-              the aircraft will contact the other controller, and you will no longer be able to
-              issue instructions to the aircraft.</p>
-            <div class="text-center">
-              <img src="outbound-handoff-comms.gif" srcset="outbound-handoff-comms-2x.gif 2x" width="133" height="79" class="img-fluid">
-            </div>
-            <br>
+              accepted the track, use the command <code>FC</code> to instruct the aircraft to contact the next controller.
+              aircraft will contact the controller that you handed off the track too. If you wish to handoff control to a seperate controller, the
```
checkandmate1, line 1740: This is the plan I have to change handoff control to make it a little more realistic. (LIBE hands directly off to area C, LIBE doesn't handoff control to LIBN for the JFK north gates).

PR #188, opened 2024-04-28
Changed frequency change to FC

### LINKED ISSUE #181: Change frequency switch to a separate command
[Discord Thread](https://discord.com/channels/1090685318014976081/1233407827913998419)

checkandmate1:
Removed "greening" to handoff control, instead `FC` should be used
Closes #181

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -1736,13 +1736,8 @@ how it simulates aircraft or the STARS interface.
             </div>
             <br>
             <p>When you are ready to transfer control of the aircraft to the controller who has
-              accepted the track, slew the aircraft's radar track. The datablock will turn green,
-              the aircraft will contact the other controller, and you will no longer be able to
-              issue instructions to the aircraft.</p>
-            <div class="text-center">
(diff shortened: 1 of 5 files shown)


mmp (commented):
Wow, that was surprisingly straightforward. Can you update:
- the `website/index.html` documentation
- `primaryAcCommands` in ui.go so that this appears in the online help when the keyboard icon in the menu bar is clicked? That uses a custom markup where * marks text to be in monospaced font and _ marks text to be italic. Then the first entry is the command, the second its description, and the third is an example of its use. (Sorry that's not better commented.)

website/index.html
```diff
@@ -1762,7 +1757,9 @@ <h3 id="stars-datablock-types">Datablock Types</h3>
               </p>
 
             <h4>Limited Datablocks (LDBs)</h4>
-
+              <p>Limited datablocks are shown when a target is not tracked by a TCP. This will contain the code the aircraft is squawking, the aircrafts
+                altitude, and when slewed, it will display the aircraft's groundspeed for a short period of time.
```
checkandmate1, line 1761: I'm not sure why this wasn't in there before, but it's now added.

mmp: The datablock documentation is a big WIP at this point, so that's the least of it's issues at the moment..

Issue #116, opened 2024-03-17
<RNGRING> keybind does not allow for numerical entry


svalencia014:
According to CRC docs, you can use the <RNGRING> (cntl + f9) to enter a numerical range ring value using the following syntax: `<RNGRING>(2,5,10,20)<ENTER>`

mmp:
Thanks for reporting. Amusingly we have the code to let you enter those values but seem to be missing the code that notices ctrl-F9 and interprets that as `<RNGRING>`. Should be an easy fix. https://github.com/mmp/vice/blob/6898b7c507cbc81e4a0e984c858e132944364802/stars.go#L2980

svalencia014:
The weird thing is using ctrl-F9 activates the range ring DCB button, it just doesn't allow numerical entry.

PR #190, opened 2024-04-30
Rework handling of keyboard input for DCB spinners

### LINKED ISSUE #107: support keyboard entry after clicking DCB
Most buttons allow keyboard input as well as using the mouse. e.g. for Range a number can be entered (and then enter pressed). We should review the docs and wire up all of the ones that offer this.

mmp:
Now there is a DCBSpinner interface and implementations of it encapsulate the sometimes-esoteric logic about which values are valid, how keyboard input is interpreted, etc. When a spinner is active (either due to clicking the DCB or due to an associated keyboard command like ctrl F9 for range rings), keyboard input from the user is passed along to set the associated value.

This also includes some updates to DCB f-key mappings to match CRC STARS. (Ctrl-F3: BRITE, Ctrl-F4: leader line length, Ctrl-F5: char size, F9: VFR plan)

Fixes #107 and #116.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -125,6 +125,7 @@ var (
 	ErrSTARSIllegalTrack      = NewSTARSError("ILL TRK")
 	ErrSTARSIllegalValue      = NewSTARSError("ILL VALUE")
 	ErrSTARSNoFlight          = NewSTARSError("NO FLIGHT")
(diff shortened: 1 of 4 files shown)


checkandmate1 (approved):
Looks good!

mmp:
Thanks!

PR #192, opened 2024-04-30
Fix regression in handling of track history

### LINKED ISSUE #64: fix wobbly PTLs when the sim rate is increased
When the sim rate is sped up, aircraft leader lines sometimes jump around between the true leader line and pointing straight up.

The underlying issue seems to be that STARS is capturing new radar tracks more quickly than the Sim is updating aircraft positions (why?). In turn, computing the heading based on the last two track positions doesn't work out since they're the same and a 0 heading leader line ends up being drawn.

A hacky workaround would be to go back and find the last radar track with a different location and use that to compute the aircraft's heading, but it would be nice to chase down the true issue.

mmp:
The updated handling of track history in 8460812c10838 inadvertently introduced a regression, where (for example) the PTLs for aircraft in a turn would only be updated at the history rate (~5s) and not every time the radar supplied us with a new track (~1s).

With this fix now we always store the most recent and the second most recent returns from the radar regardless of the track history rate used to display track tails. These are then used to calculate the aircraft heading and such, which fixes the regression...

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -383,12 +383,18 @@ const (
 )
 
 type STARSAircraftState struct {
+	// Independently of the track history, we store the most recent track
+	// from the sensor as well as the previous one. This gives us the
+	// freshest possible information for things like calculating headings,
+	// rates of altitude change, etc.
+	track         RadarTrack
+	previousTrack RadarTrack
+
 	// Radar track history is maintained with a ring buffer where
 	// historyTracksIndex is the index of the next track to be written.
 	// (Thus, historyTracksIndex==0 implies that there are no tracks.)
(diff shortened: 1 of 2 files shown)


checkandmate1:
While testing the program I noticed that #64 persists with a sim rate of 11x+.

mmp:
Rats, yes, I see that as well with 11x+. Commit message updated to remove that.

mmp:
I believe I've fixed the wobble at higher sim rates now.

checkandmate1 (approved):
Looks good! The issue with the wobble above 11x has been fixed.

Issue #191, opened 2024-04-30
add support for displaying requested altitude


mmp:
6-107 and 6-108

PR #193, opened 2024-05-01
STARS: add support for displaying requested altitude in the FDB

mmp:
6-107 / 6-108. Fixes issue #191.

### DIFF
--- a/stars.go
+++ b/stars.go
@@ -392,8 +392,9 @@ type STARSAircraftState struct {
 	historyTracks      [10]RadarTrack
 	historyTracksIndex int
 
-	DatablockType DatablockType
-	FullLDB       time.Time // If the LDB displays the groundspeed. When to stop
+	DatablockType            DatablockType
+	FullLDB                  time.Time // If the LDB displays the groundspeed. When to stop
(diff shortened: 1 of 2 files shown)


stars.go
```diff
@@ -3956,17 +3975,34 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
...
 					} else {
 						state.DisplayPTL = !state.DisplayPTL
 						status.clear = true
 					}
-				} else {
-					status.err = ErrSTARSCommandFormat
+					return
+				case "A": // toggle requested altitude: 6-108
+					if state.DisplayRequestedAltitude == nil {
+						b := ps.DisplayRequestedAltitude // inherit from system-wide
```
checkandmate1, line 3989: I can't find any of the error checking/ handling being here. I think a simple check if the data block is an FDB is all that is needed. (If it isn't, `ILLFNCT` should be shown.

mmp: Good catch, fixed.

checkandmate1 (approved):
Looks good.

PR #196, opened 2024-05-07
STARS font update

mmp:
image


radarContacto:
Nice! FYI, in the picture at top of page, the outline of characters appears offset/embossed. The char outlines should not be offset. 
image

mmp:
@radarContacto thanks for pointing that out! And now I understand what the "outline" versions of the STARS fonts are for. Fixed now for tracks; there are probably other places that should be updated similarly...

image

PR #202, opened 2024-05-10
Fix sequencing challenge

### LINKED ISSUE #201: Fix sequencing challenge
It's been really bugging me when there are multiple of the same departure gates in a row, so this will add a cap to the amount of same-gate departures

checkandmate1:
Fixed the sequencing challenge by adding a cap to the amount of same-gate departures.

Fixes #201

### DIFF
--- a/world.go
+++ b/world.go
@@ -924,7 +924,7 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
 		dep = &ap.Departures[idx]
 	}
 
-	if lastDeparture == nil || (dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap) {
+	if lastDeparture != nil || (dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap) {
 		return nil, nil, fmt.Errorf("couldn't make a departure")
 	}
 
--- a/world.go
+++ b/world.go
@@ -888,7 +888,7 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
 
 	var dep *Departure
 	if w.sameDepartureCap == 0 {
-		w.sameDepartureCap = 3
+		w.sameDepartureCap = rand.Intn(3) + 1 // Set the initial max same departure cap (1-3)
 	}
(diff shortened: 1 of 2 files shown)


world.go
```diff
@@ -916,6 +940,20 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
...
 	}
+	for {
+		if lastDeparture != nil && dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap {
+			_, dep, _ = getExit()
+		} else {
+			break
+		}
+	}
+
+	if w.sameGateDepartures == w.sameDepartureCap+4 || (lastDeparture != nil && dep.Exit != lastDeparture.Exit) { // reset back to zero if its at 7 or if there is a new gate
```
mmp, line 951: I'm a little unclear on why the `+4` here? More generally, the `3` and `4` constant should probably be named, e.g. `const maxSameGateDepartures = 3`, etc, just to make the intent clear.

world.go
```diff
@@ -917,6 +924,20 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
 		dep = &ap.Departures[idx]
 	}
 
+	if lastDeparture == nil || (dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap) {
```
mmp, line 927: Is the `lastDeparture == nil` test correct? (Should this be `if lastDeparture != nil && dep.Exit == ...`?) Currently doesn't this mean we never spawn a departure since `lastDeparture` is `nil` the first time this is called and so it will never return a valid departure?

world.go
```diff
@@ -924,7 +924,7 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
 		dep = &ap.Departures[idx]
 	}
 
-	if lastDeparture == nil || (dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap) {
+	if lastDeparture != nil || (dep.Exit == lastDeparture.Exit && w.sameGateDepartures >= w.sameDepartureCap) {
```
mmp, line 927: Shouldn't the `||` be an `&&`?

PR #211, opened 2024-05-13
WIP: Improve Flight Plan Processing

checkandmate1:
This will be the PR for the improved flight plan processing. The plan is to improve how flight plans are associated with aircraft. Instead of having `NewFlightPlan()` associate the flight plans, STARS will look at the BCN codes and flight plans to see if any of them match, and if they do, associate the track with the flight plan (Which will show the callsign, arrival airport, etc.)

Another plan for the PR is to build up the STARS-ERAM infrastructure. In the World struct, there is a `map[string]EramComputer`. Each `EramComputer` struct will contain all stored flight plans, and `map[string]STARSComputer`. Each STARSComputer will receive messages and flight plans from the ERAM computer, and local flight plans.

More will be added as the PR progresses.


processing.go
```diff
@@ -0,0 +1,885 @@
...
+	ContainedPlans    map[Squawk]*STARSFlightPlan
+	RecievedMessages  []FlightPlanMessage
+	TrackInformation  map[string]*TrackInformation
+	ERAMInbox         *[]FlightPlanMessage // The address of the overlying ERAM's message inbox.
+	Identifier        string
+	STARSInbox        map[string]*[]FlightPlanMessage // Other STARS Facilities inbox.
+	UnsupportedTracks map[int]*UnsupportedTrack
+}
+
+type STARSFlightPlan struct {
```
mmp, line 455: Something to consider (no change requested, though feel free...): should these fields just be added to `FlightPlan`? i.e., does vice have a need for a `FlightPlan` that doesn't have an associated `STARSFlightPlan`?

processing.go
```diff
@@ -0,0 +1,885 @@
...
+	FlightPlan        *STARSFlightPlan
+	PointOut          string
+	PointOutHistory   []string
+	RedirectedHandoff RedirectedHandoff
+}
+
+func (comp *STARSComputer) SendTrackInfo(receivingFacility string, msg FlightPlanMessage, simTime time.Time, Type int) {
+
+	msg.MessageType = Type
+	msg.SourceID = fmt.Sprintf("%v%v", comp.Identifier, simTime.Format("1504Z"))
```
mmp, line 530: `msg.SourceID = comp.Identifier + simTime.Format("1504Z")`?

processing.go
```diff
@@ -0,0 +1,885 @@
...
+	if comp.ContainedPlans == nil {
+		comp.ContainedPlans = make(map[Squawk]*STARSFlightPlan)
+	}
+	if comp.TrackInformation == nil {
+		comp.TrackInformation = make(map[string]*TrackInformation)
+	}
+	for _, msg := range comp.RecievedMessages {
+		switch msg.MessageType {
+		case Plan:
+			sq, _ := ParseSquawk("0000")
```
mmp, line 673: `Squawk(0)` as above..

processing.go
```diff
@@ -0,0 +1,885 @@
...
+			valid := true
+			for _, digit := range fmt.Sprintf("%04d", squawk) {
+				if digit > '7' {
+					valid = false
+					break
+				}
+			}
+
+			for _, plan := range comp.FlightPlans {
+				if plan == nil || plan.AssignedSquawk == squawk {
```
mmp, line 766: Admittedly I don't know the circumstances under which `plan == nil` but this seems strange to reject the squawk code over (especially since it will keep hitting after a new squawk is generated.) Is something like `if plan != nil && plan.AssignedSquawk == squawk` intended?

processing.go
```diff
@@ -0,0 +1,885 @@
...
+			if valid {
+				return squawk
+			}
+		}
+	}
+}
+
+// For local codes
+func (comp *STARSComputer) CreateSquawk(x int) Squawk {
+	for {
```
mmp, line 785: (Similar comments to the above function)

processing.go
```diff
@@ -0,0 +1,885 @@
...
+		badCodes := []Squawk{0o1200, 0o7500, 0o7600, 0o7700, 0o7777}
+		if slices.Contains(badCodes, squawk) {
+			continue
+		}
+		// Check if the squawk ends in 00 or is 0000.
+		if squawk%100 != 0 && squawk != 0 {
+			// Check if any digit is greater than 7.
+			valid := true
+			for _, digit := range fmt.Sprintf("%04d", squawk) {
+				if digit > '7' {
```
mmp, line 759: Although this will hit, I think it's ok / the test isn't right. In other words, your assignment to `squawk` above should indeed only generate (octal) squawk codes between 1001 and 7777, but here, you're printing them in base 10. (So for example, the octal 3410 will come out as 1800 here.) If you used (IIRC) `%04o`, I think this test would never hit. (So I think the whole test can be dropped.)

processing.go
```diff
@@ -0,0 +1,885 @@
...
+				delete(comp.TrackInformation, msg.Identifier)
+			}
+		}
+	}
+	clear(comp.RecievedMessages)
+}
+
+// For NAS codes
+func (comp *ERAMComputer) CreateSquawk() Squawk {
+	for {
```
mmp, line 747: Infinite `for` loops like this make me nervous since a) they may never terminate, in which case vice would just hang and b) in this case, when there are few squawk codes available, the loop may run many times, with each one looping over every flight plan and every `TrackInformation` to make sure it's ok.

As an alternative, you could have `ERAMComputer` maintain a `map[int]interface{}` of available squawk codes. At initialization time you can seed that with everything except `0o1200` and the other invalid squawk codes. Then if you maintain the invariant that if a squawk code is in use, it is not in that map, you just need to pick a random squawk code out of the map and delete it from the map here.

Extra bonus, since go randomizes map iteration, you could just do something like:
```
for sq := range comp.availableSquawkCodes {
  delete(comp.availableSquawkCodes, sq)
  return sq
}
```

This also would simplify error handling; you could return an error of the map was empty.

main.go
```diff
@@ -281,7 +281,7 @@ func main() {
 		var world *World
 
 		localServer = <-localSimServerChan
-
+		globalConfig.Sim = nil
```
mmp, line 284: ?

sim.go
```diff
@@ -1369,10 +1380,11 @@ func (s *Sim) GetWorldUpdate(token string, update *SimWorldUpdate) error {
 			})
 		}
 
-		var err error
+		var err error // figure out why this isn't working:
```
mmp, line 1383: Delete comment :-)

mmp (commented):
Comments from a first pass. I will continue looking but thought I'd send these along.

processing.go
```diff
@@ -0,0 +1,885 @@
...
+			return fix
+		}
+		for _, waypoint := range ac.Nav.Waypoints {
+			if waypoint.Fix == fix {
+				return fix
+			}
+		}
+
+	}
+	distanceMap := make(map[string]float32) //  -->
```
mmp, line 167: Is `distanceMap` needed? It's only used to find the closest fix and then discarded, so why not just keep track of the closest one a few lines down where you're currently setting `distanceMap[fix]`?

processing.go
checkandmate1: One of the things in `STARSFlightPlan` that's different from `FlightPlan` is that the altitude field is a string instead of an int. It could be done but given the limited amount of time, I think I should
\ do this when I come back

sim.go
```diff
@@ -1511,7 +1523,7 @@ func (s *Sim) updateState() {
 	now := s.SimTime
 
 	for callsign, t := range s.Handoffs {
```
mmp, line 1525: Let's change `t` to something like `h` or `ho` since it's a `Handoff` struct now.

sim.go
```diff
@@ -1605,19 +1674,46 @@ func (s *Sim) updateState() {
 				ac.ControllingController = ctrl
 			}
 
-			// Cull far-away departures/arrivals
+			// Cull far-away departures/arrivals. TODO: Delete these fps
+
+			deleteFromAll := func() {
```
mmp, line 1679: I think this would be cleaner to have as an ERAMComputer method, since it's digging into the internals of it. Maybe `DeleteByCallsign` and a `DeleteBySquawk` methods?

mmp: Or maybe just `DeleteAircraft(ac *Aircraft)` since I suppose the caller doesn't know which of the two will match to trigger deletion..

config.go
```diff
@@ -33,6 +33,7 @@ import (
 // 21: STARS DCB drawing changes, so system list positions changed
 // 22: draw points using triangles, remove some CommandBuffer commands
 // 23: video map format update
+// 24: flight plan processing
 const CurrentConfigVersion = 23
```
mmp, line 37: still need to increment the const

panes.go
```diff
@@ -1159,6 +1161,14 @@ func (mp *MessagesPane) processEvents(w *World) {
 				mp.input.cmd = event.Callsign + " " + cmd
 				mp.runCommands(w)
 			}
+		case AcceptedHandoffEvent, AcceptedRedirectedHandoffEvent:
+			if ac, ok := w.Aircraft[event.Callsign]; ok && event.FromController == w.Callsign {
+				if mp.NextController == nil {
+					mp.NextController = make(map[string]string)
+				}
+				mp.NextController[ac.Callsign] = event.ToController
+				fmt.Printf("%v: setting next controller for %v to be %v.\n", w.Callsign, ac.Callsign, event.ToController)
```
mmp, line 1170: Switch to `lg.Infof`

mmp (commented):
(Still getting my head wrapped around all this. A few small things as I go.)

aviation.go
```diff
@@ -1563,6 +1592,17 @@ func parseAirports() map[string]FAAAirport {
 			}
 		})
 
+	artccsRaw := LoadResource("airport_artccs.json")
+	data := make(map[string]string) // Airport -> ARTCC
+	json.Unmarshal(artccsRaw, &data)
+
+	for name, artcc := range data {
+		if entry, ok := airports[name]; ok {
+			entry.ARTCC = artcc
```
mmp, line 1601: This can be a TODO, but should there be a check in `scenario.go` (or somewhere) we have the ARTCC for the airports used in each scenario? I suppose it's not all airports, since we don't need it for an EGLL departure to KJFK, but do we need it for all of the airports that are controlled under vice in a scenario, for example?)

aviation.go
```diff
@@ -1563,6 +1592,17 @@ func parseAirports() map[string]FAAAirport {
 			}
 		})
 
+	artccsRaw := LoadResource("airport_artccs.json")
+	data := make(map[string]string) // Airport -> ARTCC
+	json.Unmarshal(artccsRaw, &data)
+
+	for name, artcc := range data {
+		if entry, ok := airports[name]; ok {
```
mmp, line 1600: I think you may want to move this block of code to `parseCIFP()`. So this function loads the airports from the ourairports.com airports database, which we mostly use for the international airports. Then `parseCIFP` loads the airports from the FAA CIFP database. In `InitializeStaticDatabase`, we merge the airports though note that the ones from `parseCIFP` take priority. So I think these `ARTCC` entries are getting discarded in that merge.

(This is all admittedly overly complicated and subtle, though is driven by wanting to load all of these databases concurrently so it gets done quickly, so vice starts quickly...)

eventstream.go
```diff
@@ -207,17 +211,19 @@ func (t EventType) String() string {
 		"OfferedHandoff", "AcceptedHandoff", "AcceptedRedirectedHandoffEvent", "CanceledHandoff", "RejectedHandoff",
 		"RadioTransmission", "StatusMessage", "ServerBroadcastMessage", "GlobalMessage",
 		"AcknowledgedPointOut", "RejectedPointOut", "Ident", "HandoffControll",
-		"SetGlobalLeaderLine", "TrackClicked"}[t]
+		"SetGlobalLeaderLine", "TrackClicked", "AutoAssociateFP", "ForceQL"}[t]
```
mmp, line 214: Need entries for `DataAcceptance` and `DataRejection`

nav.go
```diff
@@ -1419,6 +1419,31 @@ func (nav *Nav) distanceToEndOfApproach() (float32, error) {
...
 }
 
+func (nav *Nav) distanceAlongRoute(fix string) (float32, error) {
+	if nav.Heading.Assigned != nil {
+		return 0, ErrNotFlyingRoute
+	}
+	if wp := nav.Waypoints; len(wp) == 0 {
+		return 0, nil
+	} else {
+		index := -1
```
mmp, line 1429: `index := slices.IndexFunc(nav.Waypoints, func (wp Waypoint) bool { return wp.Fix == fix })`

resources/scenarios/ewr.json
```diff
@@ -3587,7 +3595,7 @@
           ]
         },
         "assigned_altitude": 6000,
-        "cruise_altitude": 39000,
+        "cruise_altitude": 6000,
```
mmp, line 3598: Is this change intended (TNCM arrival?)

resources/scenarios/ewr.json
```diff
@@ -4489,8 +4530,73 @@
     }
   },
   "stars_config": {
-  "center": "N040.43.34.389,W074.28.54.713",
-  "range": 60,
+    "center": "N040.43.34.389,W074.28.54.713",
+    "range": 60,
+    "beacon_bank": 6,
+    "coordination_fixes": {
```
mmp, line 4536: This seems redundant with the information in `adaptations.json`. Should we instead have something like

```json
"coordination_fixes": [ "FLOSI", "DYLYN", ... ]
```

in the scenario files and then issue an error if the corresponding entries aren't present in `adaptations.json` under the scenario's ARTCC?

This could be a case where it makes sense to do something like this in `STARSFacilityAdaptation`:
```go
type STARFacilityAdaptation struct {
  // ...
  CoordinationFixStrings []string  `json:"coordination_fixes"`
  CoordinationFixes   map[string]AdaptationFixes
```
So the JSON takes an array of strings and then in the `STARSFacilityAdaptation` `PostDeserialize()` method it makes sure that the entries in `CoordinationFixStrings` are all valid and then initializes `CoordinationFixes`. Then all of the code that currently uses `CoordinationFixes` can continue doing its thing the way it's written now.

scenario.go
```diff
@@ -65,6 +65,10 @@ type STARSFacilityAdaptation struct {
 	Range               float32                          `json:"range"`
 	Scratchpads         map[string]string                `json:"scratchpads"`
 	VideoMapFile        string                           `json:"video_map_file"`
+	CoordinationFixes   map[string]AdaptationFixes       `json:"coordination_fixes"`
```
mmp, line 68: Related to all of this, can you write at least some high level documentation about what the implications of this are for facility engineers? i.e., all of the new fields in JSON and roughly how they work.

scenario.go
```diff
@@ -65,6 +65,10 @@ type STARSFacilityAdaptation struct {
...
+	SingleCharAIDs      map[string]string                `json:"single_char_aids"` // Char to airport
```
mmp, line 69: We should have some code in `STARSFacilityAdaptation` `PostDeserialize()` that validates that all of the airports in this map are known to us. (In general, I've found that catching these sorts of errors at startup time is a big win in terms of making sure we get valid scenarios from people doing facility engineering; otherwise these sorts of things turn into things that look like bugs in vice that after a bunch of debugging become apparent that they're bugs in scenarios...)

scenario.go
```diff
@@ -65,6 +65,10 @@ type STARSFacilityAdaptation struct {
...
+	BeaconBank          int                              `json:"beacon_bank"`
```
mmp, line 70: Here also, validate that it's between 0-7, I suppose? (I assume that 0 is invalid but counts as not specified?)

world.go
```diff
@@ -55,6 +55,9 @@ type World struct {
 		departures map[string]map[string]map[string]bool // airport->runway->exit
 	}
 
+	ERAMComputers  map[string]*ERAMComputer
+	NonSortedPlans []*FlightPlan
```
mmp, line 59: It looks like this isn't currently used anywhere. Remove for now?

mmp (commented):
(Still working my way through it, more comments along the way.)

aviation.go
```diff
@@ -270,9 +289,12 @@ func (f FlightRules) String() string {
 }
 
 type FlightPlan struct {
+	Callsign               string
```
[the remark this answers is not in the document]
checkandmate1: Yes. This should probably be commented but I'm now treating `ac.Callsign` as sort of the ADS-B callsign for the aircraft (for the beaconator) and to avoid using the track location for the `TrackInformation` map in `STARSComputer` and `ERAMComputer`

`ac.FlightPlan` won't be used outside of `World CreateArrival`, `World CreateDeparture`, and `Nav Waypoints`.
It is used to input that flight plan information into the ERAM computer.

For example, there is an unassociated track that both N90 and PHL can see. N90 creates an abbreviated flight plan for that track with Squawk 0301 and a callsign of N1234. PHL does the same, except the callsign is N4321. If the aircraft is squawking 0301, both N90 and PHL will associate tracks and see N1234 and N4321 respectively.  In this case, both facilities are accessing the flight plan associated with that track to get the callsign.

If a controller from N90 and a controller from PHL both press F1 and look at the ADS-B callsign of the aircraft, they'll see the same callsign. I thought `ac.Callsign` was basically the ADS-B callsign just because it is global.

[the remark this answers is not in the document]
mmp: Got it. Makes sense.

processing.go
```diff
@@ -0,0 +1,1041 @@
...
+				stars.STARSInbox[fac] = address
+			}
+		}
+	}
+
+}
+
+// Access their ERAM and STARS computers. Leave blank for own TRACON
+// TODO: Change the error stuff to return an err value
+func (w *World) SafeFacility(inputTracon string) (*ERAMComputer, *STARSComputer) {
```
mmp, line 110: Why "Safe" in the name here?

world.go
```diff
@@ -943,11 +1028,33 @@ func (w *World) CreateDeparture(departureAirport, runway, category string, chall
...
+	artcc, _ := w.SafeFacility("")
+	// Add the flight plan to the ERAM computer
+
+	if artcc.FlightPlans == nil {
+		artcc.FlightPlans = map[Squawk]*STARSFlightPlan{}
+	}
+
+	starsFP := flightPlan.STARS()
+	for fix, info := range artcc.Adaptation.CoordinationFixes {
+		if strings.Contains(flightPlan.Route, fix) {
```
mmp, line 1046: Same comment as earlier about concerns about `strings.Contains` matching substrings here. Add a utility function `func RouteContainsFix(route string, fix string) bool` that then does the check more carefully and use it here and elsewhere?

processing.go
```diff
@@ -0,0 +1,1041 @@
...
+stars:
+receivesmessages
+idt
+eraminbox
+unsupported
+starsinboxes
+trackinfo
+*/
+
+func (w *World) initComputers() {
```
mmp, line 29: In general, I think it would be good to have `World` have as little knowledge of the ERAM computers and their organization as possible (and so move the code for things like this method, `SafeFacility`, `FacilityFromController`, and `UpdateComputers` out of `World` entirely.)

Maybe add a new type like `ERAMComputers` that is the one that holds the `map[string]*ERAMComputer` of all of them. Then it can have a method for looking up the `ERAMComputer` for a facility or what have you. I think that would help encapsulate a lot of logic related to handling them in a more logical place.

processing.go
```diff
@@ -0,0 +1,1041 @@
...
+	w.ERAMComputers = make(map[string]*ERAMComputer)
+	for fac := range database.ARTCCs {
+		w.ERAMComputers[fac] = &ERAMComputer{
+			Identifier:       fac,
+			ReceivedMessages: &[]FlightPlanMessage{},
+		}
+		availibleSquawks := make(map[int]interface{})
+		for i := 0o1001; i <= 0o7777; i++ {
+			availibleSquawks[i] = nil
```
mmp, line 38: Skip `0o1200`, `0u7500`, etc?

processing.go
```diff
@@ -0,0 +1,1041 @@
...
+
+		}
+		starsAvailibleSquawks := make(map[int]interface{})
+
+		bank := int64(w.STARSFacilityAdaptation.BeaconBank)
+		min, _ := strconv.ParseInt(fmt.Sprint(1+bank*100), 8, 64)
```
mmp, line 44: I think this can just be written like `1+bank*0u100` (similarly for max). (The octal stuff for beacon codes is admittedly tricky.)

sim.go
```diff
@@ -1529,6 +1541,27 @@ func (s *Sim) updateState() {
 
 			ac.TrackingController = ac.HandoffTrackController
 			ac.HandoffTrackController = ""
+			w := s.World
```
mmp, line 1544: The following ~20 lines are another case that would probably make sense as an `ERAMComputer` method in `processing.go`.

sim.go
```diff
@@ -1569,11 +1602,47 @@ func (s *Sim) updateState() {
 				s.eventStream.Post(Event{
 					Type:           OfferedHandoffEvent,
 					Callsign:       ac.Callsign,
-					FromController: ac.TrackingController,
+					FromController: ac.WaypointHandoffController,
 					ToController:   ctrl,
 				})
+				w := s.World
+				controller := w.GetControllerByCallsign(ac.TrackingController)
```
mmp, line 1609: `ERAMComputer` method?

server.go
```diff
@@ -1206,6 +1294,7 @@ func (sd *SimDispatcher) RunAircraftCommands(cmds *AircraftCommandsArgs, result
 					return nil
 				}
 			}
+			fmt.Printf("Command: %s\n", command)
```
mmp, line 1297: Delete

Issue #37, opened 2024-03-15
Add option for full-screen display


mmp:
On windows at least it apparently still shows the upper titlebar on the window: https://discord.com/channels/1090685318014976081/1216129270187819100/1216129270187819100

PR #220, opened 2024-05-19
Add option for full-screen display (#37)

excile1:
Addresses #37.

- adds an icon in the top bar to enter/exit full-screen
- adds an option in the settings to switch between different monitors (to target while full-screen) & a setting to start full-screen
- also checks if the window is already in native full-screen (macOS) and, if so, ignores the button for application full-screen


platform.go
```diff
@@ -114,7 +121,18 @@ func NewGLFWPlatform(io imgui.IO, windowSize [2]int, windowPosition [2]int, mult
 	if multisample {
 		glfw.WindowHint(glfw.Samples, 4)
 	}
-	window, err := glfw.CreateWindow(windowSize[0], windowSize[1], "vice", nil, nil)
+	var window *glfw.Window
+	if globalConfig.StartInFullScreen {
+		monitors := glfw.GetMonitors()
+		if globalConfig.FullScreenMonitor < len(monitors)-1 {
```
mmp, line 127: I think this should be `if globalConfig.FullScreenMonitor >= len(monitors)`.

I did a bunch of unfriendly tests with adding and removing an external display and found that, on a Mac laptop, if I:

- had the laptop open and plugged into an external display (so 2 monitors, the laptop screen + the external display)
- chose the external display (number 1) as the full screen display in preferences
- quit vice, closed the laptop (so 1 display now, just the external display)
- restarted vice

Then I got this crash since `globalConfig.FullScreenMonitor` was 1 and the `if` test didn't kick in...
```
panic({0x102fcbc60?, 0x14000d84450?})
	/opt/homebrew/Cellar/go/1.22.3/libexec/src/runtime/panic.go:770 +0x124
main.NewGLFWPlatform({0x18?}, {0xbc0, 0x69c}, {0xfffffffffffffcf8, 0xfffffffffffff964}, 0x0)
	/Users/mmp/vice/platform.go:131 +0x3c4
main.main()
	/Users/mmp/vice/main.go:235 +0x748
```

platform.go
```diff
@@ -154,6 +174,71 @@ func (g *GLFWPlatform) EnableVSync(sync bool) {
 	}
 }
 
+func (g *GLFWPlatform) EnableFullScreen(fullscreen bool) {
+	if g.IsMacOSNativeFullScreen() {
+		return
+	}
+
+	monitors := glfw.GetMonitors()
+	monitor := monitors[globalConfig.FullScreenMonitor]
```
mmp, line 183: I worry about the chance of a crash here as well if the active monitors change while vice is running and then `globalConfig.FullScreenMonitor` indexes past the end of the array.

(In general, I think any behavior on vice's front is fine in this case just so long as it keeps drawing a window and doesn't crash. No need to be fancy about handling this unusual case otherwise.)

mmp: (Ah, looking more, maybe this is ok and handled in `MonitorCallback` below?)

mmp (commented):
This is great--fullscreen mode is a remarkably better experience than windowed, IMHO. One crash case to take a look at and a few questions but otherwise this looks good!

mmp:
Also, please run `gofmt -w *.go` for formatting consistency.

platform.go
excile1: Apologies, it looks like I fat-fingered it while rewriting this part of the code, it should've been `if globalConfig.FullScreenMonitor > len(monitors)-1`, but I changed it to `if globalConfig.FullScreenMonitor >= len(monitors)`. I also moved the check above the `if globalConfig.StartInFullScreen {` clause since I was also able to crash it by changing the `globalConfig.FullScreenMonitor` value to something stupid and then going into the settings. This should prevent it. Thanks.

platform.go
excile1: Yes, theoretically it shouldn't happen (since monitor disconnects are handled in the callback), but I added in an extra check just in case.

mmp:
Looks great! How do you want to be listed in the credits? "Artem D"?

PR #222, opened 2024-05-21
Native Fullscreen for MacOs

checkandmate1:
I felt the fullscreen for MacOS was a bit wonky and thought that making it go into the native fullscreen instead would be much more seamless.


platform.go
```diff
@@ -225,7 +247,19 @@ func (g *GLFWPlatform) IsMacOSNativeFullScreen() bool {
 }
 
 func (g *GLFWPlatform) IsFullScreen() bool {
-	return g.window.GetMonitor() != nil || g.IsMacOSNativeFullScreen()
+	x := g.window.GetMonitor() != nil
+	if runtime.GOOS == "darwin" {
+		f := C.isNativeFullscreen(g.window.GetCocoaWindow())
+		var y bool
```
mmp, line 253: `y == f` after these tests, so `f` can be eliminated and then just `return f || x`

platform.go
```diff
@@ -225,7 +247,19 @@ func (g *GLFWPlatform) IsMacOSNativeFullScreen() bool {
...
+		if f {
+			y = true
+		} else {
+			y = false
+		}
+		return x || y
```
mmp, line 259: I'm a little unsure about the logic / goals here (and the variable names are a little obscure. Can this all just be something like:

```
if g.window.GetMonitor() != nil {
    return true
} else if runtime.GOOS == "darwin" && C.isNativeFullscreen(g.window.GetCocoaWindow()) {
    return true
} else {
    return false
}

File: platform.go [hunk 3]
```diff
@@ -7,6 +7,26 @@
 
 package main
 
+/*
+#cgo darwin CFLAGS: -x objective-c
+#cgo darwin LDFLAGS: -framework Cocoa
+
+#import <Cocoa/Cocoa.h>
```
mmp, line 14: I think you want to wrap these imports with an `#if defined(__APPLE__)` / `#endif`

platform.go
```diff
@@ -7,6 +7,26 @@
...
+#import <GLFW/glfw3.h>
+
+// Function to set macOS specific properties
+void makeFullscreenNative(void *window) {
```
mmp, line 18: Then you'll need to define these functions for all platforms, just as no-ops if not on a mac. Even though they're only called for `GOOS == "darwin"` the compiler is still probably going to want definitions for them. So more `#ifdef (__APPLE__)`/ `#else` / `#endif` should take care of that.

platform_linux.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux
```
mmp, line 1: I don't think this is needed; per https://pkg.go.dev/cmd/go#hdr-Build_constraints if the filename ends in `_linux.go` then that should be equivalent.

(Applies here and elsewhere.)

.github/workflows/ci-mac.yml
```diff
@@ -43,28 +43,26 @@ jobs:
...
-      run:
-        ./vice -lint
+      run: ./vice -lint
 
     - name: Create icons
       run: |
         mkdir -p icon.iconset
         cp icons/tower-rounded-inset-16x16.png icon.iconset/icon_16x16.png
-        cp icons/tower-rounded-inset-32x32.png icon.iconset/icon_16x16@2.png
+        cp icons/tower-rounded-inset-32x32.png icon.iconset/icon_16x16@2x.png
```
mmp, line 55: Is this intended? If so, is it fixing something?

.github/workflows/ci-mac.yml
```diff
@@ -96,7 +93,6 @@ jobs:
       env:
         APPLE_DEVELOPER_ID_CERT_PASSWORD: ${{ secrets.APPLE_DEVELOPER_ID_CERT_PASSWORD }}
         APPLE_DEVELOPER_ID_CERT_FILE: ${{ secrets.APPLE_DEVELOPER_ID_CERT_FILE }}
-      # https://medium.com/anchore-engineering/developers-need-to-handle-macos-binary-signing-how-we-automated-the-solution-part-2-ad1e08caff0f
```
mmp, line 99: I'd like to keep this in there for future reference

.github/workflows/ci-mac.yml
```diff
@@ -83,7 +81,6 @@ jobs:
         cp resources/mva-fus3.zip Vice.app/Contents/Resources/
 
     - name: Check Secrets
-      # The secrets aren't available for PRs but we don't want the build to fail just because of that...
```
mmp, line 86: Should keep this comment. More generally, the comments seem to have been deleted from this file?

platform_darwin.go
```diff
@@ -0,0 +1,553 @@
+//go:build darwin
+// +build darwin
+
+// platform.go
```
mmp, line 4: This split ends up with a lot of duplicated code across OSes which will be a pain in the future if functionality is added or if there are bugfixes in the platform code. I think a better approach is to still have a `platform.go` with all of the OS-independent stuff (the majority of the file), and to then have platform-specific files with only the things that are different on different OSes (at this point, just the implementations of `EnableFullscreen` and `IsFullscreen` for each one.)

fullscreen_darwin.go
```diff
@@ -0,0 +1,90 @@
...
+    [nswindow setCollectionBehavior: NSWindowCollectionBehaviorFullScreenPrimary];
+    [nswindow toggleFullScreen:nil];
+}
+
+// Function to check if the window is in native fullscreen mode
+int isNativeFullscreen(void *window) {
+    NSWindow *nswindow = ((NSWindow*)window);
+    return (nswindow.styleMask & NSWindowStyleMaskFullScreen) == NSWindowStyleMaskFullScreen ? 1 : 0;
+}
+#else
```
mmp, line 25: I assume the #else block isn't needed since this is only built for darwin?

fullscreen_darwin.go
```diff
@@ -0,0 +1,90 @@
...
+
+package main
+
+/*
+#ifdef __APPLE__
+#cgo darwin CFLAGS: -x objective-c
+#cgo darwin LDFLAGS: -framework Cocoa
+
+#include <Cocoa/Cocoa.h>
+#include <GLFW/glfw3.h>
```
mmp, line 11: I think this #include isn't needed.

fullscreen_darwin.go
```diff
@@ -0,0 +1,90 @@
...
+import "C"
+
+import (
+	"runtime"
+
+	"github.com/go-gl/glfw/v3.3/glfw"
+)
+
+func (g *GLFWPlatform) EnableFullScreen(fullscreen bool) {
+	if runtime.GOOS == "darwin" {
```
mmp, line 49: This test will always be true -> I think this function can be simplified to just the following two lines and then everything else deleted.

fullscreen_darwin.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		g.window.SetMonitor(nil, globalConfig.InitialWindowPosition[0], globalConfig.InitialWindowPosition[1], windowSize[0], windowSize[1], glfw.DontCare)
+	}
+}
+
+func (g *GLFWPlatform) IsFullScreen() bool {
+	if g.window.GetMonitor() != nil {
+		return true
+	} else if runtime.GOOS == "darwin" && C.isNativeFullscreen(g.window.GetCocoaWindow()) == 1 {
```
mmp, line 85: Similarly we know that `runtime.GOOS == "darwin"` for all of the code in this file...

fullscreen_linux.go
```diff
@@ -0,0 +1,45 @@
...
+package main
+
+import (
+	"runtime"
+
+	"github.com/go-gl/glfw/v3.3/glfw"
+)
+
+func (g *GLFWPlatform) IsFullScreen() bool {
+	return g.window.GetMonitor() != nil || g.IsMacOSNativeFullScreen()
```
mmp, line 12: Is there any reason to call `g.IsMacOSNativeFullScreen()` (throughout this file and also the windows one)?

fullscreen_linux.go
```diff
@@ -0,0 +1,45 @@
...
+
+	monitor := monitors[globalConfig.FullScreenMonitor]
+	vm := monitor.GetVideoMode()
+	if fullscreen {
+		g.window.SetMonitor(monitor, 0, 0, vm.Width, vm.Height, vm.RefreshRate)
+	} else {
+		windowSize := [2]int{globalConfig.InitialWindowSize[0], globalConfig.InitialWindowSize[1]}
+
+		if windowSize[0] == 0 || windowSize[1] == 0 || runtime.GOOS == "darwin" {
+			if runtime.GOOS == "windows" {
```
mmp, line 34: The two `runtime.GOOS` tests here will always be false for linux -> can simplify the logic and the following code..

fullscreen_windows.go
```diff
@@ -0,0 +1,45 @@
...
+package main
+
+import (
+	"runtime"
+
+	"github.com/go-gl/glfw/v3.3/glfw"
+)
+
+func (g *GLFWPlatform) IsFullScreen() bool {
+	return g.window.GetMonitor() != nil || g.IsMacOSNativeFullScreen()
```
mmp, line 12: `g.IsMacOSNativeFullScreen()` (here and below)

fullscreen_windows.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	monitor := monitors[globalConfig.FullScreenMonitor]
+	vm := monitor.GetVideoMode()
+	if fullscreen {
+		g.window.SetMonitor(monitor, 0, 0, vm.Width, vm.Height, vm.RefreshRate)
+	} else {
+		windowSize := [2]int{globalConfig.InitialWindowSize[0], globalConfig.InitialWindowSize[1]}
+
+		if windowSize[0] == 0 || windowSize[1] == 0 || runtime.GOOS == "darwin" {
```
mmp, line 33: GOOS will never be darwin and will always be "windows" -> can simplify the logic/code...

Issue #214, opened 2024-05-16
Redirect handoff fixes


mmp:
[discord](https://discord.com/channels/1090685318014976081/1240518110604427355/1240518110604427355)

How it should be:

1) You can't redirect to an interfacility position (outside your local STARS system) - ILL TRK. Right now on Vice, I can redirect to an external facility. 

2) You can't redirect to the controller who initiated the handoff - ILL POS.

3) If redirected, the position symbol of who the handoff has been redirected to should display. I keep seeing "R" an "P" on Vice when that is not the position id of whom it has been redirected. 

4) When redirected handoff is accepted, "RD" only remains at the original local handoff initiator for around 30sec or until clicked. "RD" is removed immediately  from tag at the position that accepted the handoff and the position that redirected the handoff.

5) If your position is not involved in a redirect (if you're not the originator, redirector, or redirected receiver), you will not see "RD" in the tag and it will apear as a normal handoff to you.

Note: This is site adaptable so might be slightly different between facilites. E.g., some facilites allow redirect but do not show the indicator. The above description is how it works at most places though.

excile1:
Since #223 was merged, this issue can be closed @mmp

PR #223, opened 2024-05-26
Redirect handoff fixes

excile1:
Addresses #214.

- prohibits inter-facility redirects (which should return a ILL TRK error), prohibits redirecting to the controller who initiated the handoff (which should return a ILL POS error)
- only shows "RD" to the original owner, last redirector and the redirect target
- "RD" is removed after the RD is accepted for everyone, except the original owner, who will see "RD" for about ~30 seconds or until clicked 
- shows sectorID as a handoff target in the datablock, instead of just a P/R
- when a RD is accepted, flashes the datablock to the original owner and makes them acknowledge it (similar to normal handoffs)
- only allows the latest redirector to recall the RD
- forces FDBs for all the redirectors even after the RD is accepted
- if the handoff itself is cancelled, clears the RD
- if the 2nd redirector RDs back to the 1st redirector forces a PDB, clears the RD to show it as a normal handoff


eventstream.go
```diff
@@ -186,6 +186,7 @@ const (
 	PointOutEvent
 	OfferedHandoffEvent
 	AcceptedHandoffEvent
+	AcceptedRedirectedHandoffEvent
```
mmp, line 189: You'll also want to update String() below to account for this.

This is also a chance that will be backwards incompatible with the server, since when events are passed over RPC the indexing will be different (and also because now there is a new event type that older versions don't handle.) So, please bump up `ViceRPCVersion` in `server.go` to account for this.

platform.go
```diff
@@ -117,6 +117,8 @@ func NewGLFWPlatform(io imgui.IO, windowSize [2]int, windowPosition [2]int, mult
 
 	// Start with an invisible window so that we can position it first
 	glfw.WindowHint(glfw.Visible, 0)
+	// Disable AutoIconify to stop the window from automatically minimizing in full-screen
```
mmp, line 120: Intended for a different PR?

stars.go
```diff
@@ -1408,6 +1410,19 @@ func (sp *STARSPane) processEvents(w *World) {
 				}
 			}
 
+		case AcceptedRedirectedHandoffEvent:
+			state, ok := sp.Aircraft[event.Callsign]
+			if !ok {
```
mmp, line 1415: We should be careful to not access state in the code a few lines below in this case. (Though, I can't remember the last time there was a bug such that we had an `*Aircraft` where the `STARSPane` hadn't created the corresponding `STARSAircraftState`; the logic around all this is pretty straightforward, so there may be an argument for not being so paranoid about this...)

stars.go
```diff
@@ -3840,6 +3863,13 @@ func (sp *STARSPane) executeSTARSClickedCommand(ctx *PaneContext, cmd string, mo
 				control := sp.lookupControllerForId(ctx, cmd, ac.Callsign)
 				if control != nil {
 					if ac.HandoffTrackController == ctx.world.Callsign || ac.RedirectedHandoff.RedirectedTo == ctx.world.Callsign { // Redirect
+						if (len(ac.RedirectedHandoff.Redirector) == 1 || (len(ac.RedirectedHandoff.Redirector) > 1) && ac.RedirectedHandoff.Redirector[1] == ctx.world.Callsign) && ac.RedirectedHandoff.Redirector[0] == control.Callsign {
```
mmp, line 3866: This logic seems to come up a few times. I'm wondering if `RedirectedHandoff` should get a few methods to encapsulate these sorts of checks so that for example here on the STARS side wouldn't so much be digging into that struct.

mmp (commented):
I'll look again more carefully tomorrow but here are a few initial comments/suggestions.

eventstream.go
excile1: Good point, done!

platform.go
excile1: I wanted to sneak that in, since it's just a one line change, but I've reverted it and will open a separate one

stars.go
excile1: I wrapped everything in `AcceptedRedirectedHandoffEvent` with the `if` clause, similarly to how other events do it, so, that should be better? Not sure if there's something else I could change?

stars.go
excile1: Done, I've added in 4 methods: `GetLastRedirector`, `ShowRDIndicator`, `AddRedirector`, `ShouldFallbackToHandoff`, which should reduce the repetitiveness of the checks and make it a bit more readable.

aircraft.go
```diff
@@ -610,3 +610,34 @@ func (ac *Aircraft) AircraftPerformance() AircraftPerformance {
...
+	} else {
+		return ""
+	}
+}
+
+func (rd *RedirectedHandoff) ShowRDIndicator(callsign string, RDIndicatorEnd time.Time) bool {
+	// Show "RD" to the redirect target, last redirector until the RD is accepted.
+	// Show "RD" to the original owner up to 30 seconds after the RD is accepted.
+	return rd.RedirectedTo == callsign || rd.GetLastRedirector() == callsign ||
+		rd.OriginalOwner == callsign || time.Until(RDIndicatorEnd) > 0
```
mmp, line 629: Should this be `(rd.OriginalOwner == callsign && time.Until(RDIndicatorEnd) > 0)`? From the comment above it sounds like the 30 seconds is just for the original owner.

mmp (commented):
One question but otherwise looks good to me. It'd be great if @checkandmate1 has a chance take a quick look as well since he wrote most of the related code originally.

aircraft.go
excile1: It doesn't have to be, since the `RDIndicatorEnd` is set in the STARS pane aircraft state, and currently there's just one check in `AcceptedRedirectedHandoffEvent` to only set that for the original owner.

mmp:
Looks great, merged!

PR #239, opened 2024-06-02
Add S56 and BIL, update ZLC maps

jlefkoff:
Adds initial scenarios for S56 (SLC) and BIL. No airspace definitions yet. Updates ZLC map file with new additions from RW data import.


mmp:
Looks great, thanks!

Issue #231, opened 2024-06-01
restart fails after quitting when in full-screen mode but minimized


mmp:
[discord](https://discord.com/channels/1090685318014976081/1090685496411295804/1246396148219121674)

PR #240, opened 2024-06-04
Stop window from creating out of bounds, disable GLFW_AUTO_ICONIFY

excile1:
Addresses #231 (and also fixes the window minimizing automatically while in fullscreen)

- adds a check to prevent negative or out of bounds `InitialWindowPosition` values (for example, if you close vice while it's minimized on Windows, the `InitialWindowPosition` would be `[-32000, -32000]`)
- disables `GLFW_AUTO_ICONIFY` to stop it from automatically minimizing


mmp:
Looks good! Nice to have that bug fixed.

PR #250, opened 2024-06-16
Paste into messages pane

checkandmate1:


panes.go
```diff
@@ -963,12 +968,24 @@ func (mp *MessagesPane) processKeyboard(ctx *PaneContext) {
 			mp.input.cursor = len(mp.input.cmd)
 		}
 	}
+	cmd := false
+	window := glfw.GetCurrentContext()
```
mmp, line 972: In general, I'd like to keep glfw stuff isolated in platform.go and hidden behind the Platform `interface`, just to have some degree of abstraction on top of all this stuff. So if this check is necessary, can you hide it behind a `Platform` method?

That said, doesn't `imgui.KeyV` already include a test that the modifier keys are pressed? (I've never used it, so that might be wrong, but that was my sense of it.)

Issue #271, opened 2024-06-29
add 'say altitude' and 'say heading' commands


mmp:
related, [discord](https://discord.com/channels/1090685318014976081/1254096432231878709/1254096432231878709)

PR #273, opened 2024-06-29
Add 'say altitude' and 'say heading' commands 

jessie846:
Fixes https://github.com/mmp/vice/issues/271


nav.go
```diff
@@ -1747,6 +1747,45 @@ func (nav *Nav) SaySpeed() PilotResponse {
...
+}
+
+func (nav *Nav) SayAltitude() PilotResponse {
+	currentAltitude := nav.FlightState.Altitude
+	var output string
+
+	if nav.Altitude.Assigned != nil {
+		assignedAltitude := *nav.Altitude.Assigned
+		if assignedAltitude < currentAltitude {
+			output = Sample(fmt.Sprintf("at %s feet descending to %s feet", FormatAltitude(currentAltitude), FormatAltitude(assignedAltitude)),
```
mmp, line 1775: Please remove "feet" from the responses since it shouldn't be there if the aircraft is in the flight levels (in which case `FormatAltitude` will return something like "FL200"). (And I think it's fine to not have "feet" below the flight levels.)

mmp (commented):
Looks great; one fix and it's good to go.

nav.go
jessie846: Will do! Thank you

jessie846:
That should fix it ^

mmp:
Looks good, thanks!

PR #279, opened 2024-07-10
add R90

golfyplayz:
add R90 with Omaha, Lincoln, and Offutt. Logan S. did most of this, I just added some finishing touches and reformatted some stuff for updates.


mmp:
Thanks!

PR #285, opened 2024-07-21
add error for FC on owned tracks

### LINKED ISSUE #230: FC on an owned track doesn't give an error
instead, we get a readback that they're switching to our frequency (and a "seeya" or whatever)

yourcptainspeaking:
Fixes #230 

Add an error message when attempting to use the "FC" command on an owned track. This checks if the new frequency is the same as the current frequency and, if so, the pilot will respond with "Unable, we are already on [frequency]"

### DIFF
--- a/pkg/sim/sim.go
+++ b/pkg/sim/sim.go
@@ -2261,7 +2261,7 @@ func (s *Sim) HandoffControl(token, callsign string) error {
 					radioTransmissions = append(radioTransmissions, av.RadioTransmission{
 						Controller: ac.ControllingController,
 						Message:    "Unable, we are already on " + octrl.Frequency.String(),
-						Type:       av.RadioTransmissionUnexpected,
+						Type:       av.RadioTransmissionReadback,
 					})
 					return radioTransmissions
 				}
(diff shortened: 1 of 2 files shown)


pkg/sim/sim.go
```diff
@@ -2257,6 +2257,14 @@ func (s *Sim) HandoffControl(token, callsign string) error {
 		func(ctrl *av.Controller, ac *av.Aircraft) []av.RadioTransmission {
 			var radioTransmissions []av.RadioTransmission
 			if octrl := s.State.Controllers[ac.TrackingController]; octrl != nil {
+				if octrl.Frequency == ctrl.Frequency {
+					radioTransmissions = append(radioTransmissions, av.RadioTransmission{
+						Controller: ac.ControllingController,
+						Message:    "Unable, we are already on " + octrl.Frequency.String(),
+						Type:       av.RadioTransmissionUnexpected,
```
mmp, line 2264: This should be `av.RadioTransmissionReadback` but otherwise this looks great; thanks for taking care of it.

(Basically RadioTransmissionReadback is for a response to a controller instruction, while RadioTransmissionUnexpected is for an aircraft calling in without an instruction right before--so things like contact after a handoff.)

mmp (commented):
And go ahead and add yourself to the credits around line 1020 in `ui.go` while you're at it..

yourcptainspeaking:
Fixed now, I thought `RadioTransmissionUnexpected` was for situations like an instruction being invalid for the aircraft.

PR #311, opened 2024-08-09
Update ewr.json

Mike-LeGall:
Line 2580 Changed "R4" to "R04"
Scratchpad won't except  2 digit input.


mmp:
It was actually a (recently-fixed) vice bug that it wasn't allowing those scratchpad inputs. Can you check if `R4` works in the latest beta for you?

Mike-LeGall:
It works now, Going to remove the leading '0' from KMMU and KTEB

PR #358, opened 2024-09-03
Added Approach/Departure airspace to EWR

Mike-LeGall:
Added approach and departure airspace to ewr and its satellites.

### DIFF
--- a/resources/openscope-aircraft.json
+++ b/resources/openscope-aircraft.json
@@ -1,6 +1,5335 @@
 {
   "aircraft": [
   {
+    "name": "Aero Commander 500",
+    "icao": "AC50",
+    "engines": {
+      "number": 2,
+      "type": "P"
+    },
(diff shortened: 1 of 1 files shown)


Mike-LeGall (commented):
I don't  know how this got added to the pull rquest. The change made to the tools.go file. I revoved it from my branch after the build. Haven't been able to build vice from scratch on my pc yet . Made a small change for 3 reasons,
- needed to make a change to get github to build the vice.exe.
- wanted to see if the created vice.exe file for windows would run on my pc, that worked.
- see if reducing the lower/upper display to 1 digit would look better and see if command still functions. Haven't had the time to investigate that yet. 
What did I do wrong, how can I create a vice.exe file for testing on github that doesn't get pulled to the main branch? I was going to try and delete it, but had second thoughts about it.

PR #391, opened 2024-10-06
add coordination lists for austin tracon

yourcptainspeaking:
Adds coordination lists for the 2 satellites (GTU and HYI) in the Austin TRACON. Also combines the two separate "stars_config" objects that were in the file for some reason


mmp:
Looks good, thanks!

PR #395, opened 2024-10-12
Add OJTI

### LINKED ISSUE #290: add instructor mode
Similar to observer, but allow issuing commands to any aircraft

[discord](https://discord.com/channels/1090685318014976081/1267319355746816030/1267319355746816030)

checkandmate1:
Fixes #290

### DIFF
--- a/pkg/sim/manager.go
+++ b/pkg/sim/manager.go
@@ -54,7 +54,7 @@ func (sm *SimManager) New(config *NewSimConfiguration, result *NewSimResult) err
 	if config.NewSimType == NewSimCreateLocal || config.NewSimType == NewSimCreateRemote {
 		sim := NewSim(*config, sm.scenarioGroups, config.NewSimType == NewSimCreateLocal, sm.mapManifests, sm.lg)
 		sim.prespawn()
-		return sm.Add(sim, result, config.Instructor)
+		return sm.Add(sim, result)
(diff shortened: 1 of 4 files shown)


pkg/sim/manager.go
```diff
@@ -86,7 +86,7 @@ func (sm *SimManager) New(config *NewSimConfiguration, result *NewSimResult) err
 	}
 }
 
-func (sm *SimManager) Add(sim *Sim, result *NewSimResult) error {
+func (sm *SimManager) Add(sim *Sim, result *NewSimResult, instuctor bool) error {
```
mmp, line 89: Interestingly, this change breaks loading of saved local sim sessions (with a "Error loading local sim: rpc: can't find method SimManager.Add callstack="[{File:main.go Line:236 Function:main}]"" error at startup time.)

The issue is that SimManager Add() is called via RPC by way of the call to `ConnectionManager` `LoadLocalSim()` from `main.go`. (It is also called purely server side, without RPC, when a remote sim is created.) The RPC library we're using requires that all RPC methods have a function signature with just two values, one for input and one for the output (and where the second one is a pointer so it can be written to.) Thus, the RPC call fails since there's no applicable function to call.

So anyway, this will have to be restructured to that still works.

mmp:
Looks good!

PR #411, opened 2024-10-28
Changed coding information for IF and FAF

Umoxfo:
The Final Approach Course Fix (`I`) may be the Intermediate Fix (IF). 
The Final Approach Fix (`F`) is established as a Final Approach Fix (FAF).


mmp:
Thank you!

mmp:
Also, @Umoxfo how would you like to be listed in the credits. "Makoto Sakaguchi"?

Issue #186, opened 2024-04-28
CIFP: Invalid output if the altitude restriction is a range


Umoxfo:
Since the ARINC 424 specification and the FAA CIFP amendment do not define the order of the altitude fields, it should not be assumed that they are always described in a specific order.

Example:
```
PS C:\Program Files (x86)\Vice> ./vice.exe -routes KSMF
PS C:\Program Files (x86)\Vice> STARs:
[...]
SLMMR5.DONNR: DONNR CARVV/a22000-20000 SLMMR/a17000-13000/s280 POZUR/a8000+
SLMMR5.WEBGO: WEBGO/a22000-20000 SLMMR/a17000-13000/s280 POZUR/a8000+
SLMMR5.ANAHO: ANAHO WEBGO/a22000-20000 SLMMR/a17000-13000/s280 POZUR/a8000+
SLMMR5.CARVV: CARVV/a22000-20000 SLMMR/a17000-13000/s280 POZUR/a8000+
SLMMR5.CRASY: CRASY FIORE CARVV/a22000-20000 SLMMR/a17000-13000/s280 POZUR/a8000+
SLMMR5.RWY17B: POZUR/a8000+ TENCO/a4000/flyover/h255
SLMMR5.RWY35B: POZUR/a8000+ RICEE PKAAY/a6000+/s210 THANN BANZA/a4000/flyover/h168
SUUTR4.NURAY: NURAY KAELA/a28000+ SUUTR/a26000-24000/s280 FOOLZ/a21000 GLDDD/a19000- WRAPS/a17000- PNNEN/a16000-12000 SLUIC
SUUTR4.RWY17B: SLUIC AMADR/a9000 NGGET/a9000-7000 CHNEL/a5000+ OVOME/a4000/s210/flyover/h348
[...]
```

mmp:
Thanks for reporting! Fixed now.

PR #414, opened 2024-10-30
ARINC424 parser: Apply the defined altitude order

Umoxfo:
Issue #186 states

> The order of the altitude fields is not defined.

However, the ARINC 424 specification states that the higher value will always appear first if “at or above altitude to at or below altitude” is specified.


mmp:
Thanks!

PR #423, opened 2024-11-07
Add Aviation Weather Data API client

Umoxfo:
Differences from `checkandmate1/AirportWeatherData`

* Query one or more airports at once
* Use standard library only

They also fixed the goroutine leak.

### DIFF
--- a/pkg/sim/sim.go
+++ b/pkg/sim/sim.go
@@ -16,6 +16,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	av "github.com/mmp/vice/pkg/aviation"
@@ -27,15 +28,14 @@ import (
 	"github.com/mmp/vice/pkg/util"
 
 	"github.com/brunoga/deep"
-	getweather "github.com/checkandmate1/AirportWeatherData"
 	"github.com/mmp/imgui-go/v4"
 )
 
 const initialSimSeconds = 45
 
 var (
-	airportWind = make(map[string]av.Wind)
-	windRequest = make(map[string]chan getweather.MetarData)
+	airportWind sync.Map
+	windRequest = make(map[string]chan struct{})
 )
 
 type Configuration struct {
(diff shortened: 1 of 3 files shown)


pkg/sim/sim.go
```diff
@@ -787,45 +788,61 @@ func (c *NewSimConfiguration) DrawRatesUI(p platform.Platform) bool {
...
+		}(done, airport)
+
 		return av.Wind{}, false
 	}
+}
+
+func refreshWeather() {
+	var wg sync.WaitGroup
+
+	// Wait for all active requests to complete.
```
mmp, line 833: Is there a risk that vice hangs here? Since `refreshWeather()` doesn't run asynchronously, if the weather fetches never return (e.g. because the machine is offline), then does this never return?

pkg/sim/state.go
```diff
@@ -174,44 +174,42 @@ func newState(selectedSplit string, liveWeather bool, isLocal bool, s *Sim, sg *
...
-		} else if dir == -1 {
-			wind = fmt.Sprintf("VRB%vKT", spd)
-		} else {
-			wind = fmt.Sprintf("%03d%02d", dir, spd)
-			if weather.Wgst > 5 {
-				wind += fmt.Sprintf("G%02d", weather.Wgst)
+		for i := range metar {
+			dir := metar[i].GetWindDirection()
+
+			var wind string
```
mmp, line 186: How about if we make this a `METAR` method that returns a wind string lives in `weather.go`? Then the detail that `dir == vrb` means variable can be localized there without callers having to worry about it.

mmp (commented):
Looks good overall; just two questions.

pkg/sim/sim.go
```diff
@@ -647,35 +647,36 @@ func (c *NewSimConfiguration) DrawUI(p platform.Platform) bool {
...
+				imgui.Text(fmt.Sprintf("%s at %d gust %d", dir, wind.Speed, wind.Gust))
 			} else {
-				imgui.Text(fmt.Sprintf("%v at %d", dir, wind.Speed))
+				imgui.Text(fmt.Sprintf("%s at %d", dir, wind.Speed))
 			}
+
 			uiStartDisable(!c.LiveWeather)
 			refresh := imgui.Button("Refresh Weather")
 			if refresh {
-				clear(airportWind)
```
Umoxfo, line 673: Ref. 1

pkg/sim/sim.go
```diff
@@ -647,35 +647,36 @@ func (c *NewSimConfiguration) DrawUI(p platform.Platform) bool {
...
-				imgui.Text(fmt.Sprintf("%v at %d gust %d", dir, wind.Speed, wind.Gust))
+				imgui.Text(fmt.Sprintf("%s at %d gust %d", dir, wind.Speed, wind.Gust))
 			} else {
-				imgui.Text(fmt.Sprintf("%v at %d", dir, wind.Speed))
+				imgui.Text(fmt.Sprintf("%s at %d", dir, wind.Speed))
 			}
+
 			uiStartDisable(!c.LiveWeather)
 			refresh := imgui.Button("Refresh Weather")
 			if refresh {
```
Umoxfo, line 675: Ref. 1

pkg/sim/sim.go
Umoxfo: No, whether the weather request was a success, this function will be completed.

Note: The connection timeout is 30 seconds.

pkg/sim/state.go
Umoxfo: I agree and **add** a method that returns the wind direction/speed in METAR text format.

mmp:
Looks good!

PR #427, opened 2024-11-18
aviation: Improved display of STAR routes

Umoxfo:
* Fixed potential gap
  The maximum length of a STAR Computer Code is 13. Ex. BLUMS5.RWY30R

* Display routes in ordered
  Routes are displayed in ascending order of transition name (or runway number) by transition type (en-route, runway, in that order).


mmp:
Thanks!

PR #430, opened 2024-11-22
Refactoring Windows installer

Umoxfo:
* Migrate `makeinstaller.bat` to WiX SDK-style project
* Upgrade WiX v5
* Delete all files in the `resources/scenarios` directory every install
  Addresses the issue of old files not being deleted, but will also be deleted if the user has placed their files.
  The solution is to specify the scenario directory using command line options, such as `-scenario`.
* Allow the user to choose an installation directory


mmp:
This is wonderful to have improved since the old approach was very fragile and hacked up. I have a few questions about how the version is updated. I'm clearly not a WiX expert, so apologies if any of the questions don't make sense!

Umoxfo:
If that means a new release of Vice, create a version tag in `git` as before.

Please take a look at the following link for managing the WiX project.
https://wixtoolset.org/docs/intro/#msbuild
Note: The project file is `windows/Vice-Installer.wixproj`, and the main source is `windows/Package.wxs`.

mmp:
Yes, but isn't version 0.11.6 hard-coded in `Vice-Installer.wixproj`?

.github/workflows/ci-windows.yml
```diff
@@ -69,13 +69,19 @@ jobs:
...
-      run: |
-        .\windows\makeinstaller.bat
+    - name: Install WiX
+      run: dotnet tool install --global wix
+
+    - name: Get Latest Git Tag
+      run: Write-Output ("VERSION=" + ((git describe --tags --abbrev=0) -replace 'v', '')) >> "$env:GITHUB_ENV"
+
+    - name: Build Installer
+      run: dotnet build ./windows/Vice-Installer.wixproj -r win-x64 -c Release -p:DefineConstants=Version="$env:VERSION" -o out
```
Umoxfo, line 84: Given the actual version number.

Umoxfo:
It's only for testing. The hardcoded version number is overwritten by arguments.

mmp:
Got it, looks good and thank you for those explanations!

PR #445, opened 2024-12-27
Fix go version for windows

svalencia014:
The `go build` and `go run` commands refuse to work on windows due to the project requesting version `1.23`. This PR simply modifies that to be `1.23.0`.


mmp:
Thanks!

PR #575, opened 2025-04-17
add vfr aircraft to AUS

yourcptainspeaking:
Adds VFR aircraft to the Austin TRACON scenario satellite airports. Rates are 8 per hour for KHYI and 6 per hour for KGTU


mmp:
Looks good!

PR #576, opened 2025-04-18
Display XXX for high climb rates(mmp#451) and Fix for alt-numpad crashing vice (mmp#573)

np123-glitch:
If an aircrafts climb rate is greater than 8,400 FPM the altitude part of the datablock becomes XXX and stays that way for 5 radar hits even after the aircrafts rate of climb is less than 8,400 FPM.


pkg/panes/stars/track.go
```diff
@@ -520,6 +528,32 @@ func (sp *STARSPane) updateRadarTracks(ctx *panes.Context) {
...
+func (sp *STARSPane) checkUnreasonableModeC(state *AircraftState, callsign string) {
+	changeInAltitude := float64(state.previousTrack.Altitude - state.track.Altitude)
+	changeInTime := state.previousTrack.Time.Sub(state.track.Time)
+	changeInTimeSeconds := changeInTime.Seconds()
+
+	var change float64
+
+	if changeInTimeSeconds != 0 {
+		change = changeInAltitude / changeInTimeSeconds
+	} else {
```
mmp, line 540: Since go zero-initializes by default, you can just remove the `else` block here entirely.

pkg/platform/glfw.go
```diff
@@ -422,6 +422,20 @@ func (g *glfwPlatform) keyChange(window *glfw.Window, key glfw.Key, scancode int
 }
 
 func (g *glfwPlatform) charChange(window *glfw.Window, char rune) {
+	// Ignore alt + numpad
```
mmp, line 425: Can you say more about what this check is doing? It's unclear from the code what these magic constants represent and why they are being ignored.

mmp (commented):
Two small comments from a first check. I will look at this again tomorrow and may have another comment or two but overall this looks good!

pkg/platform/glfw.go
np123-glitch: Basically whenever you do alt+any number on the numpad, it results in non-ascii characters that probably crashed vice. These are the rune codes for those rune characters. I basically searched up the rune codes for them and whenever I encounter these characters I just return the function.

pkg/panes/stars/track.go
np123-glitch: Alright I changed it!

pkg/platform/glfw.go
mmp: Let's remove this for now; I'd rather figure out exactly what the cause of that crash is in case there's a deeper bug that's worth fixing. (Also on the principle of keeping it to one change per PR.)

mmp (commented):
The unreasonable mode c stuff all looks good to me, so I'll merge this once the `charChange` stuff is backed out.

np123-glitch:
Changes done to ```charChange```

PR #579, opened 2025-04-22
Better way to see if Sim is paused

EthanMali:
Now when the sim is paused, it displays text saying "Simulation Paused" on a red background
image


mmp:
Nice! The code looks great (and nice work figuring out vice's sort of weird drawing code with the Builders and CommandBuffers.) My one request would be to have the message positioned at the top of the screen to more closely match the "failure banner" in rw STARS:

image

EthanMali:
This good?
image

Or do you want an outline like in the photo?

mmp:
Looks great!

PR #580, opened 2025-04-23
Added the Functionality to toggle the messages visibility on/off

EthanMali:
Added the Functionality to toggle the messages visibility on/off
image

image

### DIFF
--- a/pkg/panes/messages.go
+++ b/pkg/panes/messages.go
@@ -58,7 +58,7 @@ func init() {
 
 func NewMessagesPane() *MessagesPane {
 	return &MessagesPane{
-		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 15},
+		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 16},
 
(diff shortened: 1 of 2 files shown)


EthanMali:
Don't know, but you think the sound effects should be disabled when the message pane is hidden?

pkg/panes/messages.go
```diff
@@ -57,13 +58,15 @@ func init() {
 
 func NewMessagesPane() *MessagesPane {
 	return &MessagesPane{
-		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 16},
+		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 15},
```
EthanMali, line 61: That was an accident

mmp: Can you revert and push a commit with that restored?

mmp:
> Do you think the sound effects should be disabled when the message pane is hidden?

Interesting question. Probably yes? I think the use case for this would be an RPO sort of scenario where a human is entering the control commands as an instructor and then doing the readbacks themselves. So the sound effects are probably just a distraction.

PR #591, opened 2025-04-30
Allows annotating to the 3x3 grid on the right (mmp#92)

np123-glitch:


pkg/panes/flightstrip.go
```diff
@@ -146,10 +152,21 @@ func (fsp *FlightStripPane) ResetSim(client *server.ControlClient, ss sim.State,
 	fsp.addedPlans = make(map[sim.ACID]interface{})
 	fsp.CIDs = make(map[sim.ACID]int)
 	fsp.AllocatedCIDs = make(map[int]interface{})
+	fsp.annotations = make(map[sim.ACID][9]string)
 }
 
 func (fsp *FlightStripPane) CanTakeKeyboardFocus() bool { return false /*true*/ }
 
+func clamp(x, min, max int) int {
```
mmp, line 160: You can get rid of this and replace the `clamp()` call below with a call to `math.Clamp`

pkg/panes/flightstrip.go
```diff
@@ -533,6 +556,27 @@ func (fsp *FlightStripPane) Draw(ctx *Context, cb *renderer.CommandBuffer) {
...
+				if row >= scrollOffset && row < len(fsp.strips) {
+					// take focus
+					ctx.KeyboardFocus.Take(fsp)
+					fsp.selectedStrip = row
+					// column index 0..2
+					col := int((mx - annotationStartX) / widthAnn)
+					// row within the 3 rows of that cell
+					innerRow := 2 - (int(my)%int(stripHeight))/(int(stripHeight)/3)
+					ai := innerRow*3 + col
+					fsp.selectedAnnotation = clamp(ai, 0, 8)
```
mmp, line 575: Out of curiosity, why is the `clamp` necessary? It seems like it should always give a value in that range at least in principle.

If it is needed, would it be better to individually clamp `col` and `innerRow`?

pkg/panes/flightstrip.go
```diff
@@ -47,6 +47,7 @@ type FlightStripPane struct {
 	selectedStrip       int
 	selectedAnnotation  int
 	annotationCursorPos int
+	annotations         map[sim.ACID][9]string
```
mmp, line 50: What was the motivation for switching to storing them here rather than in `FlightStrip.Annoatations`? The `Annotations` in `FlightStrip` should probably be deleted in this change if they are no longer being used.

mmp (commented):
Looks good overall! A few followup questions.

pkg/panes/flightstrip.go
np123-glitch: Clamp makes sure the final annotation index AI is always between 0 and 8, so the program does not crash if the numbers go out of range. Normally col and innerRow should stay in the range 0–2, but small changes in layout or rounding errors might push one of them outside that range. You could also clamp col and innerRow separately to be more precise, but using a single clamp call at the end keeps the code simpler and still safe.

pkg/panes/flightstrip.go
np123-glitch: Rgr, will do

pkg/panes/flightstrip.go
np123-glitch: I thought that all UI things should stay separate from the core flight strip things but now that I think about it, in the future if we implement handing off of flightstrips, the annotations will be needed. I will move it back to FlightStrip.Annotations and commit again.

np123-glitch:
Hey Matt. I made some changes that should hopefully close this PR.

mmp (commented):
Looks good, but one last thing: can you capitalize `annotations` in `FlightStripPane`? When vice exits and it saves state, only capitalized fields are saved in the JSON (due to how go handles public vs internal data via capitalization). Otherwise if you quit vice and restart the annotations will be lost.

np123-glitch:
Done.

mmp:
Looks good, thanks!

PR #594, opened 2025-05-07
Add option to hide messages pane

EthanMali:
Add option to hide messages pane

I ran into a bug where hiding the pane stops the transmission audio as expected, but when I show the pane again, it plays all the transmissions that were missed while it was hidden. Right now, the only way I’ve found to avoid this is to automatically uncheck the radio transmission option when the pane is hidden.

image

### DIFF
--- a/pkg/panes/messages.go
+++ b/pkg/panes/messages.go
@@ -58,7 +58,7 @@ func init() {
 
 func NewMessagesPane() *MessagesPane {
 	return &MessagesPane{
-		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 15},
+		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 16},
 
 		HideMessagePane: false,
 	}
--- a/pkg/panes/messages.go
+++ b/pkg/panes/messages.go
@@ -40,6 +40,8 @@ type MessagesPane struct {
 	ContactTransmissionsAlert  bool
 	ReadbackTransmissionsAlert bool
 	HideMessagePane            bool
+	prevContactAlerts          bool
+	prevReadbackAlerts         bool
 
 	font            *renderer.Font
 	scrollbar       *ScrollBar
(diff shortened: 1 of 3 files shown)


pkg/panes/messages.go
```diff
@@ -58,12 +61,14 @@ func init() {
 func NewMessagesPane() *MessagesPane {
 	return &MessagesPane{
 		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 16},
+
+		HideMessagePane: false,
```
mmp, line 65: This shouldn't be needed since go zero-initializes things by default.

pkg/panes/messages.go
```diff
@@ -181,6 +209,7 @@ func (msg *Message) Color() renderer.RGB {
 }
 
 func (mp *MessagesPane) processEvents(ctx *Context) {
+
```
mmp, line 212: Please remove extra blank line.

pkg/panes/stars/commands.go
```diff
@@ -96,7 +96,11 @@ func (c CommandMode) PreviewString() string {
 	case CommandModeMin:
 		return "MIN"
 	case CommandModeTargetGen:
-		return "TG"
+		if !TargetGenLock {
```
mmp, line 99: It looks like this PR also includes a few other changes? I can handle this when merging it but if you use a separate git branch for each PR that should solve this issue.

pkg/panes/messages.go
```diff
@@ -39,6 +39,9 @@ type MessagesPane struct {
 	AudioAlertSelection        string
 	ContactTransmissionsAlert  bool
 	ReadbackTransmissionsAlert bool
+	HideMessagePane            bool
+	prevContactAlerts          bool
```
mmp, line 43: Interesting issue with the audio-I hadn't thought of that! It might be a little cleaner to now store the previous state here but instead to modify the code in `processEvents` so also check that `HideMessagePlane` isn't set before the calls to `PlayAudioOnce`.

(If there's a reason to leave it as is, then these two `prev*` member variables should both be capitalized so that they are saved/restored if vice is quite and restarted-this is due to how go handles internal vs public data with JSON serialization and the like.)

mmp (commented):
Looks good overall-a few small things.

PR #595, opened 2025-05-09
Added TGT GEN Lock via ```Left Shift``` & ```Tab```

EthanMali:
Added TGT GEN Lock via ```Left Shift``` & ```Tab```

image

### DIFF
--- a/pkg/panes/messages.go
+++ b/pkg/panes/messages.go
@@ -58,7 +58,7 @@ func init() {
 
 func NewMessagesPane() *MessagesPane {
 	return &MessagesPane{
-		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 15},
+		FontIdentifier: renderer.FontIdentifier{Name: "Inconsolata Condensed Regular", Size: 16},
 
 		HideMessagePane: false,
 	}
--- a/pkg/panes/messages.go
+++ b/pkg/panes/messages.go
@@ -40,6 +40,8 @@ type MessagesPane struct {
 	ContactTransmissionsAlert  bool
 	ReadbackTransmissionsAlert bool
 	HideMessagePane            bool
+	prevContactAlerts          bool
+	prevReadbackAlerts         bool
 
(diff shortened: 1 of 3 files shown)


pkg/panes/messages.go
```diff
@@ -39,6 +39,9 @@ type MessagesPane struct {
 	AudioAlertSelection        string
 	ContactTransmissionsAlert  bool
 	ReadbackTransmissionsAlert bool
+	HideMessagePane            bool
```
mmp, line 42: (Same issue as in the other one with multiple changes in this PR; again I can tidy it up when merging.)

pkg/panes/stars/commands.go
```diff
@@ -160,6 +164,8 @@ type CommandStatus struct {
 	err    error
 }
 
+var TargetGenLock bool
```
mmp, line 167: This should probably be a member variable in `STARSPane`. In this case I would not capitalize it, since it's something that probably shouldn't persist across quitting and relaunching vice.

pkg/panes/stars/commands.go
```diff
@@ -2313,11 +2327,14 @@ func (sp *STARSPane) executeSTARSCommand(ctx *panes.Context, cmd string, tracks
 		if trk != nil {
 			sp.runAircraftCommands(ctx, trk.ADSBCallsign, cmds)
 			status.clear = true
+
```
mmp, line 2330: Please remove added blank lines (here and 2x below)

pkg/panes/stars/commands.go
```diff
@@ -96,7 +96,11 @@ func (c CommandMode) PreviewString() string {
 	case CommandModeMin:
 		return "MIN"
 	case CommandModeTargetGen:
-		return "TG"
+		if !TargetGenLock {
```
mmp, line 99: Can you also add a sentence or two of documentation for this in the "ATC Commands" section in website/index.html?

mmp (commented):
Also looks good overall and I think will be super useful. A few comments/requests.

PR #596, opened 2025-05-11
TGT GEN Lock

EthanMali:
I separated the branches, to make it easier to merge.

image

### DIFF
--- a/pkg/panes/stars/commands.go
+++ b/pkg/panes/stars/commands.go
@@ -31,8 +31,6 @@ var pausedAircraftInfo *util.TransientMap[av.ADSBCallsign, string] = util.NewTra
 
 type CommandMode int
 
-var targetGenLock bool
-
 const (
 	// Keyboard command entry modes; can be main or DCB menu for these; sp.dcbShowAux decides.
 	CommandModeNone CommandMode = iota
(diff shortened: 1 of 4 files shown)


pkg/panes/stars/commands.go
```diff
@@ -31,6 +31,8 @@ var pausedAircraftInfo *util.TransientMap[av.ADSBCallsign, string] = util.NewTra
 
 type CommandMode int
 
+var targetGenLock bool
```
mmp, line 34: Can you make this a member variable in `STARSPane`? Otherwise looks good!

mmp:
It looks like the current version isn't building--missing an update to that call in lists.go?

> Error: pkg/panes/stars/lists.go:77:13: not enough arguments in call to sp.commandMode.PreviewString
	have ()
	want (*STARSPane)

EthanMali:
I downloaded the newest version of list.go and still get the build fail

mmp:
> I downloaded the newest version of list.go and still get the build fail

Yep. I think the issue is that you changed
```go
func (c CommandMode) PreviewString() string {
```
to
```go
func (c CommandMode) PreviewString(sp *STARSPane) string {
```
and preexisting code in lists.go has a call to `PreviewString`. So that will need an update.

EthanMali:
Doesnt want to work. I compared my list.go with the one in the repo, their identical.

image

EthanMali:
Seems to pass now, except the gofmt which failed because of aivation_test.go (which i didnt touch)

mmp:
Great, looks good!

PR #621, opened 2025-05-30
Fixes Issue#615

rickr:
Here’s my first pass at adding selectable SIDs STARs etc. 

The basics:
* adds an imgui picker to each menu
* changes the scenario line builders to color line builders. 
* uses the list DCB control to control the opacity. 

Alpha is not yet supported and in my admittedly ignorant view seems like it would be a bit bigger lift to implement. 

Happy to hear any and all feedback you may have.

### DIFF
--- a/pkg/panes/stars/tools.go
+++ b/pkg/panes/stars/tools.go
@@ -1084,73 +1084,8 @@ func (sp *STARSPane) drawScenarioRoutes(ctx *panes.Context, transforms ScopeTran
 		Color:          color,
 		DrawBackground: true}
 
-	// STARS
-	if sp.scopeDraw.arrivals != nil {
-		for _, name := range util.SortedMapKeys(ctx.Client.State.InboundFlows) {
-			if sp.scopeDraw.arrivals[name] == nil {
-				continue
-			}
-
-			arrivals := ctx.Client.State.InboundFlows[name].Arrivals
-			for i, arr := range arrivals {
-				if sp.scopeDraw.arrivals == nil || !sp.scopeDraw.arrivals[name][i] {
-					continue
-				}
-
-				drawWaypoints(ctx, arr.Waypoints, drawnWaypoints, transforms, td, style, ld, pd, ldr)
-
-				// Draw runway-specific waypoints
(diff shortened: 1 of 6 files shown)


pkg/panes/stars/tools.go
```diff
@@ -1192,25 +1254,21 @@ func (sp *STARSPane) drawScenarioRoutes(ctx *panes.Context, transforms ScopeTran
...
 		}
 	}
+	generate_commands(cb, transforms, ctx, ld, pd, td, ldr)
+}
 
-	// And now finally update the command buffer with everything we've
-	// drawn.
-	cb.SetRGB(color)
+
+func generate_commands(cb *renderer.CommandBuffer, transforms ScopeTransformations, ctx *panes.Context,
```
mmp, line 1270: Two requests here: first to match the naming conventions elsewhere (i.e., [Camel case rather than snake case](https://stackoverflow.com/questions/17326185/what-are-the-different-kinds-of-cases), and second to use a function name that's more clear about what this is for. e.g., `generateRouteDrawingCommands`?)

mmp (commented):
This looks great! Two small requests in the comments.

pkg/renderer/builders.go
```diff
@@ -190,6 +190,11 @@ func (l *ColoredLinesDrawBuilder) AddLine(p0, p1 [2]float32, color RGB) {
 	l.color = append(l.color, color, color)
 }
 
+func (l *ColoredLinesDrawBuilder) AddLineWithAlpha(p0, p1 [2]float32, color RGB) {
```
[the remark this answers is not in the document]
rickr: Good catch. 72b0e178a2ff28c49ecdd957de4181caad5e3e75

pkg/panes/stars/tools.go
rickr: Good call on the naming. fe3afe9ba50b602ebef1e60ae8c4d227b0ed6d25

rickr:
gofmt fixed in 0342c723c520e6e3c1280bcb4d698b49355cee16

rickr (commented):
gofmt job fixed with 0342c723c520e6e3c1280bcb4d698b49355cee16

PR #629, opened 2025-06-06
Modify TAF url to use new faa v3 json source

rickr:
It looks like v2 has been deprecated and v3 is current.

The summary page can be found at 
https://tfr.faa.gov/tfr3/export/json

*however* this is not an actual JSON file. 
Viewing the network requests we find the an XHR request to the URL we are using 
https://tfr.faa.gov/tfrapi/exportTfrList

This gives us a list of NOTAM IDs. 

We then assume the URL pattern for the details XML (eg https://tfr.faa.gov/download/detail_5_0205.xml).

### DIFF
--- a/pkg/aviation/db.go
+++ b/pkg/aviation/db.go
@@ -9,6 +9,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/csv"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


mmp:
Looks good!

PR #652, opened 2025-06-14
Re-add EWR Multiplayer 

checkandmate1:
Multiplayer scenarios for KEWR were accidentally removed in commit 79633de. Re-added here.


mmp:
Looks good, thanks!

PR #656, opened 2025-06-17
Departure controller for specific departure routes

checkandmate1:
Added a custom departure controller for departure routes. 

Most useful for background departures while controlling a singular sector. (Modeled in `ewr.json`)

### DIFF
--- a/server/scenario.go
+++ b/server/scenario.go
@@ -259,7 +259,7 @@ func (s *scenario) PostDeserialize(sg *scenarioGroup, e *util.ErrorLogger, manif
 							if _, ok := sg.ControlPositions[route.DepartureController]; !ok {
 								e.ErrorString("controller %q for departure route %q is unknown", route.DepartureController, fix)
 							} else if !slices.Contains(s.VirtualControllers, route.DepartureController) {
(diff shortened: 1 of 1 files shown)


pkg/server/scenario.go
```diff
@@ -250,6 +250,13 @@ func (s *scenario) PostDeserialize(sg *scenarioGroup, e *util.ErrorLogger, manif
 						if activeAirportRunways[rwy.Airport] == nil {
 							activeAirportRunways[rwy.Airport] = make(map[string]interface{})
 						}
+						if route.DepartureController != "" {
+							if _, ok := sg.ControlPositions[route.DepartureController]; !ok {
+								e.ErrorString("controller %q for departure route %q is unknown", route.DepartureController, fix)
+							} else if !slices.Contains(s.VirtualControllers, route.DepartureController) {
```
mmp, line 256: Should an error be issued if this is used for a human controlled position?

mmp (commented):
(Sorry for not getting to this before the release.)

Overall looks good; can you also add documentation in `website/index.html`?

PR #658, opened 2025-06-19
VFR <200 under Bravo shelf 

np123-glitch:


pkg/math/core.go
```diff
@@ -103,11 +107,28 @@ func Clamp[T constraints.Ordered](x T, low T, high T) T {
 	return x
 }
 
+// Lerp performs linear interpolation between a and b using factor x in [0,1]
 func Lerp(x, a, b float32) float32 {
 	return (1-x)*a + x*b
 }
 
-// greatest common divisor
+// Min returns the smaller of two float32 values.
```
mmp, line 115: Min/Max can be deleted (they were removed in 5f7954a8ff56504b56f199ae79f73783d99fe73f). Instead you can use `min()` and `max()` which were added to the core go language a little while ago.

pkg/aviation/db.go
```diff
@@ -1272,6 +1272,17 @@ func InBravoAirspace(p math.Point2LL, alt int) bool {
 	return inAirspace(DB.BravoAirspace, p, alt)
 }
 
+func UnderBravoShelf(p math.Point2LL, alt int) bool {
+	for _, vols := range DB.BravoAirspace {
```
mmp, line 1276: My main feedback is about performance here: I was wondering about the overhead of this and did some measurements with the go CPU profiler. (In vice, if you run `vice.exe -cpuprofile cpu.prof` then after you exit vice you'll have a cpu.prof file waiting for you. Then `go tool pprof vice.exe cpu.prof` gets you in the profiler.

I did a run with the 22s at JFK (where there are many VFRs and there's some bravo airspace) and played for a minute or two. Looking at `Nav.Update`, in my run it ran for 0.33s, 0.09s of which were in `UnderBravoShelf`, so about 30% of the total time. It turns out that in the worst case (which is also the common case, when `false` is returned, the two loops run 369 times to check all bravo airspace rings. Over a few tens of VFR aircraft, that adds up. (Put another way, the airspeed update code takes overall about 9x longer than the heading and altitude update code whereas before all three were about the same.)

So, it would be nice to make this faster, especially so that it doesn't risk bogging down the multicontroller server when multiple sims are running. In `aviation/db.go` there is an `AirspaceGrid` class for making these sorts of queries faster; it breaks the world into ~6nm squares and then figures out which airspace volumes overlap each square. Then when an aircraft checks airspace, it only needs to check the ones in the square it's inside.

So first I'd suggest adding a `Below` method to `AirspaceVolume`. Next, it turns out that `Sim` already maintains an `AirspaceVolume` for bravo airspace (used when figuring out if random VFR routes are ok), in `Sim.bravoAirspace`. Then pass that into `Nav` Update` and down to where `UnderBravoShelf` is currently called. That should give much better performance (and avoid risks of being a bottleneck in the future.)

mmp (commented):
Awesome! Two requests (one minor, one a little more involved but hopefully not too bad.)

np123-glitch:
Changed the code to adhere to your comments.

mmp (commented):
Looks good!

PR #659, opened 2025-06-21
fix assigned alt

checkandmate1:
If `assigned_altitude` was never set in an inbound route, the assigned altitude in the arrival struct would be zero, despite if crossing restrictions were in the route. This PR addresses that issue.

### DIFF
--- a/pkg/aviation/aviation.go
+++ b/pkg/aviation/aviation.go
@@ -769,13 +769,16 @@ func (ar *Arrival) PostDeserialize(loc Locator, nmPerLongitude float32, magnetic
 	if ar.AssignedAltitude == 0 {
 		// check the route to see altitude restrictions
 		wps := []Waypoint(ar.Waypoints)
+		altitudeRestriction := false 
 		for _, wp := range wps {
 			if wp.AltitudeRestriction != nil {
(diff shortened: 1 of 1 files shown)


mmp (commented):
Seems reasonable to issue an error if "assigned_altitude" isn't set and there are no altitude restrictions in the STAR, but I'm not sure about setting `ar.AssignedAltitude` if it's unset if there are altitude restrictions: that is more or less equivalent to a controller saying "descend and maintain (altitude)" and overrides altitude restrictions in the STAR. (See also nav.go, `Nav.TargetAltitude`, where `if nav.Altitude.Assigned != nil` is one of the first checks.) Have you tested that case in practice?

(Also, does this lead to errors being issued in any existing scenarios / does this potentially fix any currently-open bugs?)

checkandmate1:
Ok. This was originally a fix to an ERAM bug: the assigned altitude in the DB would show `000` for the arrivals that didn't have it explicitly set in the JSON. 

I also found out that multiple scenarios don't have an assigned altitude or altitude restrictions, as the arrivals come in at the initial altitude. 

I think this PR can be closed if the error isn't necessary.

PR #662, opened 2025-06-23
ERAM

checkandmate1:
MVP release:

What's done so far:
- Toolbar 
- Tracks
- Datablocks
- Font 
- Accept handoff, QU lines, leader line redirection, J rings, interim alts, assigned alts.
- Handoff
- QU Direct

~~What still needs to be completed:~~ (Not in MVP release)
- ~~Point out~~
- ~~AR list/ AR command~~


pkg/server/dispatcher.go
```diff
@@ -477,6 +477,41 @@ func (sd *dispatcher) DeleteAircraft(da *DeleteAircraftListArgs, update *sim.Sta
 	}
 }
 
+type QULineArgs struct {
+	ControllerToken string
+	ACID            sim.ACID
+}
+
+func (sd *dispatcher) SendCoordinateInfo(da *QULineArgs, update *sim.StateUpdate) error {
```
mmp, line 485: Naming inconsistency? `SendCoordinateInfo` for the method vs `QULineArgs` for the arguments?

pkg/sim/spawn.go
```diff
@@ -1098,6 +1114,25 @@ func (s *Sim) createArrivalNoLock(group string, arrivalAirport string) (*Aircraf
...
+				continue
+			}
+			if int(wp.AltitudeRestriction.TargetAltitude(arr.InitialAltitude)) < lowestAlt {
+				lowestAlt = int(wp.AltitudeRestriction.TargetAltitude(arr.InitialAltitude))
+			}
+		}
+		if lowestAlt != gomath.MaxInt {
+			assigned = lowestAlt
+		} else {
+			fmt.Println("Warning: no altitude restriction found for arrival", ac.ADSBCallsign)
```
mmp, line 1133: `lg.Warn`?

pkg/util/text.go
```diff
@@ -63,6 +63,55 @@ func WrapText(s string, columnLimit int, indent int, wrapAll bool) (string, int)
 	return result.String(), lines
 }
 
+func WrapTextNoSpace(s string, columnLimit int, indent int, wrapAll bool) (string, int) {
```
mmp, line 66: Since this is largely the same as `WrapText`, does it make sense to have a single function that takes a parameter to control this behavior? More generally, at this point maybe we should switch to a `TextWrapConfig` struct that holds parameters (columnLimit, indent, wrapAll, wrapNoSpace) with a `Wrap()` method. (Though feel free to ignore the second suggestion.)

pkg/sim/stars.go
```diff
@@ -449,6 +449,7 @@ type AirspaceAwareness struct {
 
 type STARSFlightPlan struct {
```
mmp, line 450: So either we should rename this to something like `NASFlightPlan` or I think there should be an ERAM-specific flight plan type. AFAIK a lot of what's in there now is STARS-specific (e.g. `Location` for where an unsupported datablock is, inhibit mode C display, etc.) and in general I think it's a good principle to encapsulate STARS-specific data to STARS code as much as possible.

I guess another question is whether the per-aircraft ERAM info here is a quote unquote flight plan (this is my ignorance about ERAM). (This is just trying to figure out what this blob of additional data should be named.)

main.go
```diff
@@ -305,3 +309,43 @@ func main() {
 		}
 	}
 }
+
```
mmp, line 312: With the way this is handled currently, if an ERAM scenario is launched and then you go back to STARS, all of the STARS settings (fonts, brightness, PTLs, etc.) get reset to their default; this is more or less due to vice assuming a fixed configuration of the `Pane` objects being drawn for the UI.

In retrospect, I think a better approach would be to directly store the `stars.STARSPane`,`eram.ERAMPane`, `panes.FlightStripPane`, and `panes.MessagesPane` as members in `ConfigNoSim`. Then those are the canonical instances of those that are stored/loaded when vice exits/is relaunched. If we also separately recorded the two split line positions from the `ConfigNoSim.DisplayRoot` hierarchy (to record adjustments the user made by dragging them), then we could no longer serialize `ConfigNoSim.DisplayRoot` (e.g., rename it to `displayRoot`). And then, finally, when launching a new sim, we could just create a new `*panes.DisplayNode` hierarchy from scratch each time a sim was launched, using either the STARS or ERAM pane for the radar pane, as appropriate.

(I'm happy to take a pass at this since it's preexisting technical debt, but feel free to yourself if you prefer.)

mmp (commented):
First pass, looks good in general!

PR #663, opened 2025-06-25
Control multi controller flows

checkandmate1:
Created a new icon in multicontroller scenarios. 

Departures and inbound flows will be labeled and can be changed.


pkg/renderer/font.go
```diff
@@ -180,6 +180,7 @@ var (
 	FontAwesomeIconRedo                = faUsedIcons["Redo"]
 	FontAwesomeIconSquare              = faUsedIcons["Square"]
 	FontAwesomeIconTrash               = faUsedIcons["Trash"]
+	FontAwesomeIconUsers               = faUsedIcons["Users"]
```
mmp, line 183: Please add an entry around line 450 of website/index.html documenting this icon in the main menu and what it does.

pkg/server/dispatcher.go
```diff
@@ -90,6 +90,31 @@ func (sd *dispatcher) SetLaunchConfig(lc *SetLaunchConfigArgs, _ *struct{}) erro
...
+	defer sd.sm.lg.CatchAndReportCrash()
+
+	if ctrl, s, ok := sd.sm.LookupController(mc.ControllerToken); !ok {
+		return ErrNoSimForControllerToken
+	} else {
+		return s.SetMultiControllers(ctrl.tcp, mc.Config)
+	}
+}
+
+func (sd *dispatcher) TakeOrReturnFlowControl(token string, _ *struct{}) error {
```
mmp, line 108: No change necessarily requested, but I've been wondering if it makes sense for launch control (and by extension, this addition) to have the whole flow of a single controller taking control. Alternatively, we might let any controller make changes to these settings and not try to limit it to one. (And then leave it to be a social issue among the people in a session to have consensus about what to do, versus sort of limiting it but not really since anyone can take control.)

So, if you want to remove the flow for somene taking control and then checking that the person who has control is making a change, I'm fine with that.

ui.go
```diff
@@ -1408,3 +1434,117 @@ func uiDrawSettingsWindow(c *client.ControlClient, config *Config, p platform.Pl
 
 	imgui.End()
 }
+
+func uiDrawMultiControllersWindow(c *client.ControlClient, eventStream *sim.EventStream, p platform.Platform) {
+	if !ui.showMultiControllers {
+		return
+	}
+
+	if ui.mcEntries == nil {
```
mmp, line 1443: Is there an issue here where `ui.mcEntries` is only set once and then persists? i.e., if the user connects to a different sim without exiting vice, will the old ones be shown?

More generally, I'm wondering if it would be better to not have `ui.mcEntries` at all but to generate `mcEntries []multiControllerEntry` from scratch into a local variable here each time this function runs. It shouldn't be much overhead and then avoids any risk of bugs due to fixed state that became stale due to external updated.

ui.go
```diff
@@ -1408,3 +1434,117 @@ func uiDrawSettingsWindow(c *client.ControlClient, config *Config, p platform.Pl
...
+
+		for i := range ui.mcEntries {
+			e := &ui.mcEntries[i]
+			imgui.TableNextRow()
+			imgui.TableNextColumn()
+			imgui.Text(e.Flow)
+			imgui.TableNextColumn()
+			imgui.Text(e.Type)
+			imgui.TableNextColumn()
+			imgui.InputTextWithHint("##"+e.Flow, "", &e.Controller, 0, nil)
```
mmp, line 1500: I was testing this with a JFK scenario and tried entering `2a` and got an error about an invalid controller. I believe that imgui has options to convert input to upper-case in text inputs, but it might be more friendly to use `imgui.BeginComboV` / `imgui.SelectableBoolV` to present a list of the valid controllers for the user to choose from.

ui.go
```diff
@@ -1408,3 +1434,117 @@ func uiDrawSettingsWindow(c *client.ControlClient, config *Config, p platform.Pl
...
+
+	canEdit := ctrl == c.State.UserTCP || c.State.AreInstructorOrRPO(c.State.UserTCP)
+	if !canEdit {
+		imgui.BeginDisabled()
+	}
+
+	if imgui.BeginTableV("mc", 3, imgui.TableFlagsBordersV|imgui.TableFlagsBordersOuterH|imgui.TableFlagsRowBg|imgui.TableFlagsSizingStretchProp, imgui.Vec2{}, 0) {
+		imgui.TableSetupColumn("Flow")
+		imgui.TableSetupColumn("Type")
+		imgui.TableSetupColumn("Controller")
```
mmp, line 1489: It would be helpful to add an additional column showing which signed-in controller the selected controller resolves to, for the (common) case when the selected controller isn't signed in. So when a single controller is signed in to a multi-controller scenario, we'd see that all positions end up being theirs, regardless of how things are set here, but then as multiple ones are signed in it would become apparent what the practical effect was of changing a particular assignment, if the assigned-to controller wasn't signed in.

mmp (commented):
Overall looks great, though a few comments to consider before this lands.

checkandmate1:
The consolidation more or less does this.

PR #667, opened 2025-06-27
fix: climb virtual controlled a/c to cruise

yourcptainspeaking:
Previously, aircraft that were entirely controlled by virtual controllers would not climb to cruise due to `DepartOnCourse` never being called for them. This fixes that by calling it for any departure handed off from one virtual controller to another, since these aircraft will likely never talk to a human controller

### DIFF
--- a/pkg/sim/control.go
+++ b/pkg/sim/control.go
@@ -647,6 +647,21 @@ func (s *Sim) handoffTrack(fp *STARSFlightPlan, toTCP string) {
 	s.Handoffs[fp.ACID] = Handoff{
 		AutoAcceptTime: s.State.SimTime.Add(time.Duration(acceptDelay) * time.Second),
 	}
+	callsign := func() av.ADSBCallsign {
+		// ugly way to do this but using s.CallsignForACID here causes vice to crash
+		for _, aircraft := range s.Aircraft {
+			if aircraft.STARSFlightPlan.ACID == fp.ACID {
+				return aircraft.ADSBCallsign
+			}
+		}
+		return av.ADSBCallsign("")
+	}()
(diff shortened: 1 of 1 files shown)


pkg/sim/control.go
```diff
@@ -647,6 +647,21 @@ func (s *Sim) handoffTrack(fp *STARSFlightPlan, toTCP string) {
 	s.Handoffs[fp.ACID] = Handoff{
 		AutoAcceptTime: s.State.SimTime.Add(time.Duration(acceptDelay) * time.Second),
 	}
+	callsign := func() av.ADSBCallsign {
+		// ugly way to do this but using s.CallsignForACID here causes vice to crash
```
mmp, line 651: Aha. This is poorly documented but the Sim code has the general convention that methods with lower-case names assume that the Sim.mu mutex is locked and upper-case names do not. So `CallsignForACID` tries to grab the lock but it's already held here in `handoffTrack` (and thence, presumably, that crash.)

Could you instead add a `callsignForACID` method after `CallsignForACID`, move the logic from `CallsignForACID` there, and then have `CallsignForACID` just grab the lock and call the lower-case one? Then you can use `callsignForACID` here?

Alternatively, arguably we could just punt here and do the cast `av.ADSBCallsign(aircraft.STARSFlightPlan.ACID)` and pass that to `enqueueDepartOnCourse` since it's probably fine to assume that the ACID and ADSB callsign will match for an aircraft a human hasn't interacted with; presumably those will only mismatch for things like VFRs where the human types in the wrong callsign for the ACID.

yourcptainspeaking:
Ah, thought it had something to do with the lock but I didn't look into it.
I can do that when I get home

mmp:
Looks good!

PR #669, opened 2025-07-29
Help close #628

### LINKED ISSUE #628: Sim does not close with "Start new simulation" window open
Confirmed on windows only. More data would be appreciated. 

**To repro** 
* Open the game
* Select the "Start new simulation" button on the menu.
* Click the OS' close button (eg the "X" on Windows)  to close the window.
* The close command is queued, however the window will not close until the dialogs close.

**Expected result**
The sim should close immediately when clicking on the OS' close button

np123-glitch:


mmp:
The issue with this as written is that if vice has a fatal error, we'd like to be able to show a dialog box to the user to tell them what happened, rather than just crashing. So if you add something like `ShowFatalErrorDialog(render, plat, lg, "ERROR!")` at line 394 of main.go, then without this change, it shows the error and only exits when the user clicks "ok". But with this change, vice just exits immediately without showing it.

PR #684, opened 2025-08-21
Voice to Commands

checkandmate1:
name on the tin!


cmd/vice/ui.go
```diff
@@ -1375,6 +1377,11 @@ func uiDrawSettingsWindow(c *client.ControlClient, config *Config, p platform.Pl
 		imgui.EndCombo()
 	}
 
+	// Speech-to-Text settings
```
mmp, line 1380: FWIW it was a little surprising to me that these were under STARS settings; I think having them at the same level would be worthwhile as far as being easier to find for folks.

stars/ui.go
```diff
@@ -55,6 +55,71 @@ func (sp *STARSPane) DrawUI(p platform.Platform, config *platform.Config) {
...
+		imgui.SameLine()
+		micName := ps.SelectedMicrophone
+		if micName == "" {
+			micName = "Default"
+		}
+		if imgui.BeginComboV("##microphone", micName, 0) {
+			if imgui.SelectableBoolV("Default", ps.SelectedMicrophone == "", 0, imgui.Vec2{}) {
+				ps.SelectedMicrophone = ""
+			}
+			mics := p.GetAudioInputDevices()
```
mmp, line 107: This doesn't need to be fixed if it's not straightforward, but I'm seeing some formatting issues with the text, viz:
image

stars/ui.go
```diff
@@ -423,3 +488,160 @@ func (sp *STARSPane) DrawInfo(c *client.ControlClient, p platform.Platform, lg *
 		}
 	}
 }
+
+func getKeyName(key imgui.Key) string {
```
mmp, line 492: This feels like it should live elsewhere, maybe in `platform`? Alternatively, is there an imgui function that does this? It would be nice to not have it manually done here.

stt/stt.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		return "", err
+	}
+	return command, nil
+}
+
+func Transcribe(audio *AudioData) (string, error) {
+	// Make a model
+
+	text, err := whisper.Transcribe("../../resources/models/ggml-medium-q5_0.bin", audio.Data, audio.SampleRate, audio.Channels, whisper.Options{
```
mmp, line 139: We should probably have this checked in (using git LFS, which is annoying in general but is applicable here.) Then at least it will go through the regular flow of being included in the resource file manifest, uploaded to google cloud storage for users to retrieve it the first time they run, etc.

mmp, line 139: Thoughts on getting the "finding the model file path" logic in sync with how resources are handled locally: so currently we either build vice with either `util/resources_local.go`, the default, for local development builds--like before, it looks for resources around the current directory, or `util/resources_download.go`, for user installs where it looks for resources in `%APPDATA\Vice\resources%` on windows and the equivalent on OSX.

Currently those are based on returning a `*fs.StatFS` that calling code can then use to open up files. But for the whisper APIs, we need a proper path.

So... I'd suggest changing those two files in `util` to have a public function like `GetResourcesFolderPath()` that incorporates much of the logic from the corresponding `initResourcesFS()` functions and to then have `initResourcesFS()` call that. Then here, you can do something like `modelPath := filepath.Join(util.GetResourcesFolderPath(), "models", "ggml-medium-q5_0.bin")` and then pass `modelPath` to `Transcribe`. That should make everything work as desired both for local builds and releases.

stars/ui.go
```diff
@@ -55,6 +55,71 @@ func (sp *STARSPane) DrawUI(p platform.Platform, config *platform.Config) {
 			sp.playOnce(p, i)
 		}
 	}
+
+	imgui.Separator()
+	if imgui.CollapsingHeaderBoolPtr("Speech to Text", nil) {
+		// Push-to-talk key
+		keyName := "None"
+		if ps.PushToTalkKey != imgui.KeyNone {
```
mmp, line 63: I like the idea of having a default PTT key (and then letting people change it if they don't like it) though am fine with no default if you feel strongly about it.

autowhisper/transcribe.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	return strings.TrimSpace(b.String()), nil
+}
+
+// TranscribePCM16 takes raw PCM16 samples (interleaved if stereo), automatically converts
+// them to 16 kHz mono, and returns the transcription text.
+// inSampleRate is the input sample rate in Hz, inChannels must be 1 or 2.
+func Transcribe(modelPath string, pcm []int16, inSampleRate, inChannels int, opts Options) (string, error) {
+	// Load model
+	model, err := whisper.New(modelPath)
```
mmp, line 119: I was a little worried about overhead from making a new model, context, configuring it, etc, each time `Transcribe` is called, since it seems like it could be a heavyweight operation (versus keeping the model/context around in a struct and reusing them, which I assume is an option.)

I did some quick profiling (both cpu and memory) and this doesn't seem to be an issue, so I think this is fine as is, but I figured I'd mention the thought.

autowhisper/transcribe.go
```diff
@@ -0,0 +1,266 @@
...
+	var b strings.Builder
+	for {
+		seg, err := ctx.NextSegment()
+		if err != nil {
+			if errors.Is(err, fmt.Errorf("%w", err)) { // keep linter happy; io.EOF checked below via string
+				// no-op
+			}
+		}
+		if err != nil {
+			if strings.Contains(err.Error(), "EOF") {
```
mmp, line 101: I think `errors.Is` is the idiomatic way to do this (if I'm following what you're doing here)

mmp (commented):
I know this is still a WIP but since it seems to be coming along well, here are a few comments from an initial look-over.

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+		return Segment{}, io.EOF
+	}
+	res := toSegment(c.model.ctx, c.n)
+	c.n++
+	return res, nil
+}
+
+func (c *context) IsText(t Token) bool {
+	switch {
+	case c.IsBEG(t), c.IsSOT(t), whisperlow.Token(t.Id) >= c.model.ctx.Whisper_token_eot(), c.IsPREV(t), c.IsSOLM(t), c.IsNOT(t):
```
mmp, line 132: Any reason to write it like this vs something like
```go
if c.IsBEG(t) || c.IsSOT(t) || ... {
  return false
}
return true
```
?

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+			Text:  ctx.Whisper_full_get_token_text(n, i),
+			P:     ctx.Whisper_full_get_token_p(n, i),
+			Start: time.Duration(d.T0()) * time.Millisecond * 10,
+			End:   time.Duration(d.T1()) * time.Millisecond * 10,
+		}
+	}
+	return res
+}
+
+func strconvI(i int) string { return fmt.Sprintf("%d", i) }
```
mmp, line 189: How about deleting this and replacing uses with `strconv.Itoa`?

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+		Text:   strings.TrimSpace(ctx.Whisper_full_get_segment_text(n)),
+		Start:  time.Duration(ctx.Whisper_full_get_segment_t0(n)) * time.Millisecond * 10,
+		End:    time.Duration(ctx.Whisper_full_get_segment_t1(n)) * time.Millisecond * 10,
+		Tokens: toTokens(ctx, n),
+	}
+}
+
+func toTokens(ctx *whisperlow.Context, n int) []Token {
+	res := make([]Token, ctx.Whisper_full_n_tokens(n))
+	for i := 0; i < len(res); i++ {
```
mmp, line 176: `for i := range len(res)`

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (c *context) SetLanguage(lang string) error {
+	if c.model.ctx == nil {
+		return ErrInternalAppError
+	}
+	if !c.model.IsMultilingual() {
+		return ErrModelNotMultilingual
+	}
+	if lang == "auto" {
+		c.params.SetLanguage(-1)
```
mmp, line 36: Check the returned `error`?

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+		func() bool {
+			if enc != nil {
+				return enc()
+			}
+			return true
+		},
+		func(new int) {
+			if seg != nil {
+				num := c.model.ctx.Whisper_full_n_segments()
+				s0 := num - new
```
mmp, line 102: Confirm this computation (specifically, that `num-new` is the first element processed in the `for` loop below?) I don't know the whisper.cpp API at all but from pattern matching would have guessed that `num` would be the first one processed.

autowhisper/internal/whisper/context.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (c *context) ResetTimings() { c.model.ctx.Whisper_reset_timings() }
+func (c *context) PrintTimings() { c.model.ctx.Whisper_print_timings() }
+func (c *context) SystemInfo() string {
+	return "system_info: n_threads = " +
+		strconvI(c.params.Threads()) + " / " + strconvI(runtime.NumCPU()) +
+		" | " + whisperlow.Whisper_print_system_info()
+}
+
+func (c *context) Process(data []float32, enc EncoderBeginCallback, seg SegmentCallback, prog ProgressCallback) error {
```
mmp, line 85: audioSamples (or samples) would probably be a better name than `data`.

autowhisper/internal/whisper/model.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+func (m *model) IsMultilingual() bool {
+	return m.ctx.Whisper_is_multilingual() != 0
+}
+
+func (m *model) Languages() []string {
+	result := make([]string, 0, whisperlow.Whisper_lang_max_id())
+	for i := 0; i < whisperlow.Whisper_lang_max_id(); i++ {
```
mmp, line 47: `for i := range whisperlow.Whisper_lang_max_id()`

autowhisper/transcribe.go
```diff
@@ -0,0 +1,267 @@
...
+	// Max tokens per segment (0 = no limit).
+	MaxTokensPerSegment uint
+}
+
+// TranscribeFile loads the model at modelPath, reads the WAV at audioPath,
+// converts it to 16 kHz mono if needed, and returns the full transcription text.
+//
+// Requirements:
+// - audioPath must be a WAV file. Other formats should be converted upstream.
+func TranscribeFile(modelPath, audioPath string, opts Options) (string, error) {
```
mmp, line 37: I don't think this is used anywhere. Delete?

autowhisper/transcribe.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+	}
+
+	// Resample to 16k using linear interpolation
+	const outRate = 16000
+	if inRate != outRate {
+		ratio := float64(outRate) / float64(inRate)
+		outLen := int(math.Ceil(float64(len(mono)) * ratio))
+		res := make([]float64, outLen)
+		for i := 0; i < outLen; i++ {
```
mmp, line 248: `for i := range outLen`

autowhisper/transcribe.go
```diff
@@ -0,0 +1,267 @@
...
+			}
+			mono[i] = v
+		}
+	} else { // stereo interleaved
+		if len(pcm)%2 != 0 {
+			return nil, errors.New("stereo pcm length is not even")
+		}
+		frames := len(pcm) / 2
+		mono = make([]float64, frames)
+		for i := 0; i < frames; i++ {
```
mmp, line 224: `for i := range frames`

autowhisper/wav.go
```diff
@@ -0,0 +1,86 @@
...
+	for i, v := range buf.Data {
+		fbuf.Data[i] = float64(v) / float64(1<<15)
+		if fbuf.Data[i] > 1 {
+			fbuf.Data[i] = 1
+		} else if fbuf.Data[i] < -1 {
+			fbuf.Data[i] = -1
+		}
+	}
+
+	// Mixdown to mono if stereo
```
mmp, line 50: The "mixdown to mono -> resample to 16k -> convert to float32" code here is quite similar to the corresponding code in `pcmInt16ToFloat32Mono16k`. (I haven't carefully checked all correspondences.) Refactor into a single utility function called by both if possible?

cmd/vice/ui.go
```diff
@@ -1402,6 +1406,85 @@ func uiDrawSettingsWindow(c *client.ControlClient, config *Config, p platform.Pl
...
+		imgui.Text("Push-to-Talk Key: ")
+		imgui.SameLine()
+		imgui.TextColored(imgui.Vec4{0, 1, 1, 1}, keyName)
+
+		if c.PTTCapture {
+			imgui.TextColored(imgui.Vec4{1, 1, 0, 1}, "Press any key for Push-to-Talk...")
+			if kb := p.GetKeyboard(); kb != nil {
+				for key := range kb.Pressed {
+					fmt.Println("Pressed: ", platform.GetImGuiKeyName(key))
+					if key != imgui.KeyLeftShift && key != imgui.KeyRightShift &&
```
mmp, line 1425: Are these not allowed since they don't come in the same way as regular keypresses? It might be nice to allow these (in a followup commit); I think keys like e.g. right shift are pretty useful for this sort of thing.

platform/audio.go
```diff
@@ -261,3 +262,18 @@ func audioCallback(user unsafe.Pointer, ptr *C.uint8, size C.int) {
...
+
+//export audioInputCallback
+func audioInputCallback(user unsafe.Pointer, ptr *C.uint8, size C.int) {
+	n := int(size)
+	in := unsafe.Slice(ptr, n)
+	ar := (*AudioRecorder)(user)
+
+	// Convert bytes to int16 samples
+	samples := make([]int16, n/2)
+	for i := 0; i < n/2; i++ {
```
mmp, line 274: `for i := range n/2`

platform/audio_input.go
```diff
@@ -0,0 +1,145 @@
...
+		Samples:  2048,
+		Callback: sdl.AudioCallback(C.audioInputCallback),
+		UserData: user,
+	}
+
+	// Use SDL's default device mechanism for empty string
+	var deviceID sdl.AudioDeviceID
+	var err error
+	if deviceName == "" {
+		deviceID, err = sdl.OpenAudioDevice("", true, &spec, nil, 0)
```
mmp, line 68: This is fine for now but opening the audio device each time makes me nervous about overhead. Maybe better to keep a `map[string]sdl.AudioDeviceID` in `AudioRecorder`, keyed off of `deviceName`, and then to reuse the same device each time recording is done?

platform/keymouse.go
```diff
@@ -129,6 +129,20 @@ func (g *glfwPlatform) GetKeyboard() *KeyboardState {
 	if imgui.IsKeyPressedBool(imgui.KeyInsert) {
 		keyboard.Pressed[imgui.KeyInsert] = nil
 	}
+	// For PTT
+	if imgui.IsKeyPressedBool(imgui.KeySemicolon) {
```
mmp, line 133: When I use STT (with the default semicolon for PTT), I'm seeing the below in the menubar, where my mac is letting me know that the microphone is active. It initially appears when I press `;`, disappears for a second, then reappears as long as I hold `;`. My guess is that when key repeat kicks in, it reappears. I would have expected that using `IsKeyPressedBool` like this would have meant that key repeat wasn't relevant. I have noticed that I need to wait until it's solidly on before I start talking or else it clips the start of what I'm saying.

https://github.com/user-attachments/assets/1ec0df84-7301-4334-92f6-238c63ab01fe

util/resources_local.go
```diff
@@ -10,8 +10,43 @@ package util
...
 )
 
 func initResourcesFS() *fs.StatFS {
-	return localResourcesFS()
+	dir := GetResourcesFolderPath()
+	fsys, ok := os.DirFS(dir).(fs.StatFS)
+	if !ok {
+		panic("FS from DirFS is not a StatFS?")
+	}
+
```
mmp, line 23: IMHO it's fine to `return &fsys` here without further checks. Since `GetResourcesFolderPath()` panics if it fails (and since it's right below and easy to cross-check when looking at this file), we can probably skip double-checking.

mmp (commented):
I still have to look at `stt.go` again but here are a few comments on everything else. I think all should be straightforward or just TODOs for the future.

stt/stt.go
```diff
@@ -0,0 +1,269 @@
...
+		// Independently detect release (do not tie to pressed state; key repeat may keep it true)
+		if client.PTTRecording && !imgui.IsKeyDown(PTTKey) {
+			if p.IsAudioRecording() {
+				data, err := p.StopAudioRecording()
+				client.PTTRecording = false
+				if err != nil {
+					lg.Errorf("Failed to stop audio recording: %v", err)
+				} else {
+					lg.Infof("Push-to-talk: Stopped recording, transcribing...")
+					go func(samples []int16) {
```
mmp, line 211: This I think is going to be unsafe since it accesses a bunch of things from `client` (and then calls `RunAircraftCommands` as well.) The issue is that `client.State` may be in the middle of being updated (by vice's usual `sim.GetStateUpdate` flow sending updates from the server), which could lead to panics or inconsistent results. `Airports` and `Approaches` are probably fine in practice, though not stuff like `GetTrackByCallsign`.

The broader issue is that none of the `client` stuff is written to handle concurrent access from multiple goroutines since it isn't needed elsewhere in vice so isn't worth the complexity. Client-side vice mostly runs without a lot of goroutines and most everything happens in the main thread.

So, to fix I'd suggest:
- Initialize `approaches` before launching the goroutine, while you're still running in the main thread.
- Maybe use a channel to report back `text` returned by `VoiceToCommand`.
- Then each time `ProcessSTTKeyboardInput` here is called, check the channel for a result (and then since you're in the main thread at that point it's safe to call `GetTrackByCallsign`, etc., and run the commands.)

You might need to keep a slice of channels in case another one comes in before the last one has returned from the goroutine (unlikely in general but if the network is down, etc., that could be an issue.)

stt/stt.go
```diff
@@ -0,0 +1,269 @@
...
+		text = strings.ReplaceAll(text, number, word+" ")
+	}
+	text = strings.ReplaceAll(text, "  ", " ") // Remove double spaces
+	text = strings.ToLower(text)
+	return text
+}
+
+func ProcessSTTKeyboardInput(p platform.Platform, client *client.ControlClient, lg *log.Logger,
+	PTTKey imgui.Key, SelectedMicrophone *string) {
+	if PTTKey != imgui.KeyNone {
```
mmp, line 187: Just to reduce the level of indentation here and improve comprehensibility, you could start with
```
if PTTKey == imgui.KeyNone {
    return
}
```
And then everything else is indented that much less.

mmp (commented):
Done reviewing, overall looks good but I think the race condition should be cleaned up before this lands.

PR #701, opened 2025-10-26
Parsing CT instruction after C/A/S instruction

### LINKED ISSUE #700: Bad Command Evaluation
https://github.com/mmp/vice/blob/43bccab88509674b41ec2a8c7075e36cdc90e9eb/server/dispatcher.go#L661

When trying to run the command `CTRTEL/A70/S210` to an M98 arrival on the MUSCL3 for 30R, the pilot responds with "Goodbye" in text, and "See ya" via voice. My guess is that this is just a parsing conflict that's only present for waypoints that start with `T`, resulting in the command matching with a `CT` prefix.

adamgoose:
Fixes #700

Steps to repro:
1. Load scenario M98, running straight 30s
2. Spawn an arrival on the MUSCL3
3. On contact, `DVS E30R`
4. On approaching TRTEL, `CTRTEL/A70/S210`

Before the fix, the aircraft would say "see ya". After the fix, they say "Cross TRTEL at 7,000, speed 210" correctly.

### DIFF
--- a/server/dispatcher.go
+++ b/server/dispatcher.go
@@ -658,11 +658,6 @@ func (sd *dispatcher) RunAircraftCommands(cmds *AircraftCommandsArgs, result *Ai
 					rewriteError(err)
 					return nil
 				}
-			} else if strings.HasPrefix(command, "CT") && len(command) > 2 {
-				if err := s.ContactController(tcp, sim.ACID(callsign), command[2:]); err != nil {
-					rewriteError(err)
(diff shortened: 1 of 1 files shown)


mmp:
Nice fix. I think this is the right approach (vs validating the TCP); in general I think it's most intuitive if the command syntax is unambiguous without reference to "is that a valid TCP", "is that the name of a VOR" sorts of thigns.

PR #703, opened 2025-10-27
Allow quitting while modal dialogs are open

### LINKED ISSUE #628: Sim does not close with "Start new simulation" window open

adamgoose:
Closes #628

I'm not sure why this limitation was put in place to begin with, and I'm not sure it won't have unintended consequences in other circumstances that I'm not aware of. Feedback welcome.

### DIFF
--- a/cmd/vice/main.go
+++ b/cmd/vice/main.go
@@ -480,7 +480,7 @@ func main() {
 				lg.Info("performance", "stats", stats)
 			}
 
-			if plat.ShouldStop() && len(ui.activeModalDialogs) == 0 {
+			if plat.ShouldStop() {
 				// Do this while we're still running the event loop.
 				saveSim := mgr.ClientIsLocal()
 				config.SaveIfChanged(render, plat, controlClient, saveSim, lg)


mmp:
See https://github.com/mmp/vice/pull/669 for one catch (note: with the same proposed fix). However, retesting this now, everything seems to work as expected even with fatal error dialog boxes. Could you check as well on your system, adding something like `ShowFatalErrorDialog(render, plat, lg, "oh no")` right before the `for {` loop at the bottom of main.go and confirm that the dialog is still shown/vice doesn't exit immediately in that case?

adamgoose:
However, it looks like that's an insufficient test. While trying other airports, I'm getting actual crashes, and the fatal error dialog is only visible *without* this change.

PR #759, opened 2026-01-19
feat(fe): allow specific departure routes to be hfr

yourcptainspeaking:
[discord](https://discord.com/channels/1090685318014976081/1090685459379794021/1462595416817401951)

Allows facility engineers to define specific exit routes as hold for release rather than an entire airport. If an airport is marked as HFR, then all exits will still be HFR regardless of their individual settings.


mmp:
Nice!

PR #764, opened 2026-01-20
Define-specific-callsigns

checkandmate1:
To define specific callsigns in a scenario. 

`callsigns` can be to `airlines` in lieu of `icao`.


aviation/aviation.go
```diff
@@ -203,6 +246,75 @@ func (a *AirlineSpecifier) Check(e *util.ErrorLogger) {
...
+	}
+
+	// Calculate flight distance to filter aircraft by CWT category
+	dep, arr := DB.Airports[departureAirport], DB.Airports[arrivalAirport]
+	flightDistance := math.NMDistance2LL(dep.Location, arr.Location)
+
+	// Sample according to fleet count, filtering by maximum distance for CWT category
+	var actype string
+	acCount := 0
+	for _, ac := range a.Aircraft() {
```
mmp: This and the following could use `rand.SampleWeighted` with suitable weighting functions.

website/index.html
```diff
@@ -6737,6 +6737,11 @@ <h2 class="section-heading">Airlines and Aircraft</h2>
                   <td>String</td>
                   <td>The ICAO code of the airline (which must be present in <i>vice</i>'s <code><a href="https://github.com/mmp/vice/blob/master/resources/openscope-airlines.json">openscope-airlines.json</a></code> file.)</td>
                 </tr>
+                <tr>
+                  <td>"callsign"</td>
+                  <td>String</td>
+                  <td>A specific callsign (eg. AAL123) to use for an aircraft flying this route. (Both callsigns and ICAO cannot be specified at the same time.)</td>
```
mmp: Isn't it that if a callsign is specified, then the icao is ignored? If so, should probably validate in PostDeserialize() just so users aren't confused if they give both.

sim/spawn_departures.go
```diff
@@ -615,11 +615,25 @@ func (s *Sim) createIFRDepartureNoLock(departureAirport, runway, category string
...
 
-	airline := rand.SampleSlice(s.Rand, dep.Airlines)
-	ac, acType := s.sampleAircraft(airline.AirlineSpecifier, departureAirport, dep.Destination, s.lg)
+	if len(dep.Airlines) == 0 {
+		return nil, fmt.Errorf("no airlines for departure at %q", departureAirport)
+	}
+
+	var ac *Aircraft
+	var acType string
+	for range 50 {
```
mmp: I believe that if we only have a callsign and already have an aircraft in the air with that callsign, this will end up running 50 times trying to spawn that callsign repeatedly. In that case it would be good to fail fast and move on.

checkandmate1:
Sorry for taking so long to get to this... Thanks for the feedback!

sim/spawn_arrivals.go
```diff
@@ -101,14 +102,37 @@ func (s *Sim) createArrivalNoLock(group string, arrivalAirport string) (*Aircraf
...
 	arr := arrivals[idx]
 
-	airline := rand.SampleSlice(s.Rand, arr.Airlines[arrivalAirport])
-	ac, acType := s.sampleAircraft(airline.AirlineSpecifier, airline.Airport, arrivalAirport, s.lg)
+	airlines := arr.Airlines[arrivalAirport]
+	if len(airlines) == 0 {
+		return nil, fmt.Errorf("no airlines for arrival group %s airport %s", group, arrivalAirport)
+	}
+
+	// Filter callsigns that are currently in use.
```
mmp: Sorry, one last request: the block of code here is more or less duplicated in `createOverflightNoLock()` and `createIFRDepartureNoLock()`. Move it to a helper function that those call?

PR #790, opened 2026-02-16
Refactor Scenario Definitions

checkandmate1:
- Moves `stars_config` from scenario files into each facilities' own dedicated configuration files in `resources/configurations/<artcc>/<facility>`.
   - These configuration files were derived from current scenarios.
   - For facilities that have not yet been made in *vice*, controllers were imported from CRC configuration files.
 - Configuration files have:
    - `stars_config`.
    - Neighboring facilities and their ID.
    - Area configurations (i.e., default airport for an area rather than a specific TCW). More can be added later.
    - `airspace_awareness`/ `coordination_fixes`.
    - `control_positions` for that facility.
- Removes `controllers` from scenarios. All controllers in that facility and neighboring facilities are loaded.
- Standardized the naming convention of controllers:
   - All facilities are named via their sector ID and are loaded in with how they would be referenced by that facility. 
- When naming a facility outside of that scenario's facility, `initial_facility`, `departure_facility`, or `receiving_facility` must be specified. 
- Updated `index.html` to reflect new definition changes.


resources/configurations/ZNY/N90.json
```diff
@@ -0,0 +1,2733 @@
...
+                    ],
+                    "NY_LW_APP": [
+                        "NY_LS_APP",
+                        "NY_LE_APP",
+                        "SWF_APP"
+                    ]
+                },
+                "fix_pair_assignments": [
+                    {
+                        "fix_pair_index": 0,
```
mmp: Related, is the fix_pair_index ever different than the index of the entry in the slice?

resources/saysid.json
```diff
@@ -271,6 +271,10 @@
   "LEINY": "Lenny",
   "LEJON": "Legion",
   "LGA": "LaGuardia",
+  "LGA_D_APP": "Empire",
```
mmp: I don't think this will work as intended: saysid.json is only (I believe) used to look up the way to say SIDs.

server/scenario.go
```diff
@@ -742,13 +798,8 @@ func (sg *scenarioGroup) PostDeserialize(e *util.ErrorLogger, catalogs map[strin
 			}
 			if len(ctrl.SectorID) < 2 {
 				e.ErrorString("must specify both facility and numeric sector for center controller")
-			} else {
-				if !(ctrl.SectorID[0] >= 'A' && ctrl.SectorID[0] <= 'Z') {
-					e.ErrorString("first character of center controller \"sector_id\" must be a letter")
-				}
-				if _, err := strconv.Atoi(ctrl.SectorID[1:]); err != nil {
```
mmp: Confirming, removing this was intended?

resources/configurations/ZNY/ABE.json
```diff
@@ -0,0 +1,207 @@
...
+            {
+                "id": "PHL",
+                "prefix": "3"
+            },
+            {
+                "id": "ZNY",
+                "eram_prefix": "N"
+            }
+        ],
+        "neighboring_facilities": [
```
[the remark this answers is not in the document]
checkandmate1: I originally kept it in there because I didn't have the ERAM facilities extracted yet, but now that they are, they seem to be redundant.

mmp:
Great stuff-this is a fantastic milestone.

PR #791, opened 2026-02-16
Add missing Vice logo assets for website documentation

jordw:
I noticed the logo image in the header of the Vice documentation was missing.

I created the missing PNGs at the sizes referenced in the markup from the high res source file, and it now renders as expected:
image


mmp:
Thanks!

PR #793, opened 2026-02-16
ERAM - Update History Trails/Add Radar Filter menu

jessie846:
Update History Trails to add drawing logic , 0-5.  Radar Filter --> History

Add Radar Filter Toolbar.

### DIFF
--- a/eram/cmdops.go
+++ b/eram/cmdops.go
@@ -79,6 +79,21 @@ func init() {
 	registerCommand(CommandModeNone, "MR", handleMapRequestList)
 	registerCommand(CommandModeNone, "MR [FIELD]", handleMapRequestLoad)
 
+	// LA
+	registerCommand(CommandModeNone, "LA [LOC_SYM] [LOC_SYM]", handleLALocLoc)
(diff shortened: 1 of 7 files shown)


eram/toolbar.go
```diff
@@ -510,6 +512,63 @@ func (ep *ERAMPane) drawToolbarMenu(ctx *panes.Context, scale float32) {
...
+			handleClick(ep, &ep.HistoryLength, 0, 5, 1)
+		}
+
+		p2 := [2]float32{rightEdge, toolbarDrawState.buttonCursor[1] - buttonSize(buttonFull, scale)[1]}
+		p3 := [2]float32{p0[0], p2[1]}
+
+		toolbarDrawState.lightToolbar = [4][2]float32{p0, p1, p2, p3}
+		ep.drawMenuOutline(ctx, p0, p1, p2, p3)
+
+		toolbarDrawState.customButton[fmt.Sprintf("HISTORY\n%d", ep.HistoryLength)] = toolbarButtonGreenColor
```
checkandmate1: This should be done before the button is drawn.

Also, to avoid redundant entries in `customButton`, I think it would be better if you kept this entry as `HISTORY` and then added `HISTORY` to `weirdNames` in `cleanButtonName()`

PR #816, opened 2026-02-24
Add test bench panel for easier dev testing

jordw:
As more nuanced logic gets added to the sim, I found it can be difficult to test new flows and commands thoroughly. I created a test bench panel tool to help make that easy and fun. 

Test cases are facility/scenario agnostic: placeholders like {approach}, {iaf}, {if}, {faf} resolve to actual fixes for whichever approach you've selected, so the same tests work across facilities. Initial scenarios cover approach clearance request flows and basic departure commands, but it should be straightforward to build out further. I think this can also be useful for smoke testing new facility scenarios.

The direct-placement spawn mode constructs a specially configured overflight positioned at a specific distance and bearing from a fix via LaunchArrivalOverflight, making it easy to set up precise test scenarios. Departure/arrival spawns reuse the existing CreateDeparture/CreateArrival plumbing. This should also work fine with ERAM as well, but needs a bit more work to support fully support overflights.

To try this, launch vice with the --dev flag to enable developer tools. A wrench icon appears in the toolbar - click it to open the Test Bench panel. See the attached demo for a walkthrough (commands are issued via STT so there's no audio).

I used an extended version of this to thoroughly smoke test the visual approach logic in #807, and it was quite useful. Those extra scenarios are separate for now since they depend on the visual approach logic, but you can see them here: https://github.com/jordw/vice/commit/ee904d8c3246341e638a34994376395596cc4578

https://github.com/user-attachments/assets/593c9996-924f-4c91-9c41-19b9af3fdea8

### DIFF
--- a/cmd/vice/main.go
+++ b/cmd/vice/main.go
@@ -69,6 +69,7 @@ var (
 	replayDuration    = flag.String("replay-duration", "3600", "replay `duration` in seconds or 'until:CALLSIGN'")
 	waypointCommands  = flag.String("waypoint-commands", "", "waypoint `commands` in format 'FIX:CMD CMD CMD, FIX:CMD ...,'")
 	starsRandoms      = flag.Bool("starsrandoms", false, "run STARS command fuzz testing with full UI (randomly picks a scenario)")
+	devMode           = flag.Bool("dev", false, "enable developer tools")
 )
 
 func setupSignalHandler(profiler *util.Profiler) {
--- a/cmd/vice/config.go
+++ b/cmd/vice/config.go
@@ -66,6 +66,7 @@ type ConfigNoSim struct {
 	InhibitDiscordActivity util.AtomicBool
 	NotifiedTargetGenMode  bool
 	DisableTextToSpeech    bool
+	DevMode                bool
 
 	UserWorkstation    string
 	ControllerInitials string
--- a/cmd/vice/simconfig.go
+++ b/cmd/vice/simconfig.go
@@ -1323,6 +1323,24 @@ func (c *NewSimConfiguration) DrawConfigurationUI(p platform.Platform, config *C
 	// TRAFFIC RATES section
 	drawSectionHeader("Traffic Rates")
 
+	// In developer mode, offer a quick toggle to disable all traffic.
+	if *devMode || config.DevMode {
+		lc := &c.ScenarioSpec.LaunchConfig
+		noTraffic := lc.DepartureRateScale == 0 && lc.InboundFlowRateScale == 0 &&
+			lc.VFRDepartureRateScale == 0
+		if imgui.Checkbox("No traffic (test bench)", &noTraffic) {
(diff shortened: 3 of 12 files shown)


jordw:
For arrivals, I'm using test callsigns for now, but departures use the airlines from the facility. It's easy enough to do the same with arrivals as well, as that might be better for facility testing.

jordw:
I cleaned up the code a bit, and also added scenarios for most of the other STT commands

cmd/vice/main.go
```diff
@@ -69,6 +69,7 @@ var (
 	replayDuration    = flag.String("replay-duration", "3600", "replay duration in seconds or 'until:CALLSIGN'")
 	waypointCommands  = flag.String("waypoint-commands", "", "waypoint commands in format 'FIX:CMD CMD CMD, FIX:CMD ...,'")
 	starsRandoms      = flag.Bool("starsrandoms", false, "run STARS command fuzz testing with full UI (randomly picks a scenario)")
+	devMode           = flag.Bool("dev", false, "enable developer tools")
```
mmp: For the folks who aren't command-line users (especially those doing facility engineering), let's also add a developer mode checkbox in the settings window, maybe right after "Disable text-to-speech". (There's no great place for it and that window could use a cleanup, but that's another issue.)

mmp (changes requested):
This is great and will be super helpful going forward.

Can you add some documentation about all this in the Facility Engineering section of index.html? Use of `-dev` to enable it / settings window), a general explanation of how to use it., etc. (e.g. in my first try it wasn't evident to me that I was supposed to issue the instructions myself.)

A few bugs from trying it out, all with JFK dep 13, land 13L/22L:

- FRG approach clearance requests to ILS 14 (at least) seem broken:

Image

- If I give two commands at once rather than giving them sequentially, it doesn't seem to reflect the second one in the UI (`DROSLY`) still has a > next to it:

Image

- If I spawn one for an approach clearance (JBU2388) and then spawn one for departure basics (LXJ116) then it shows the approach's callsign in the UI for the suggested command.

Image

- Possibly related, it seems to be stuck on [wait: check_in] for that departure even though we have gotten a departure contact message from it;

Image

- This is getting logged each time an aircraft is spawned; it's probably benign for these purposes but I'm curious what's going wrong with squawk code assignments here

```
time=2026-02-25T06:41:36.806-08:00 level=WARN msg="DPJ1214: squawk code \"0000\" not in any pool" sim_name=funny-border callstack="[sim.go:2059:/sim.(*Sim).CheckLeaks.func1 sim.go:2067:/sim.(*Sim).CheckLeaks sim.go:1159:/sim.(*Sim).Step sim.go:1129:/sim.(*Sim).Update manager.go:436:/server.(*SimManager).runSimUpdateLoop asm_arm64.s:1268:runtime.goexit]"
```

mmp:
(Closing after offline discussion.)

PR #817, opened 2026-02-25
Move stars_maps into area_configs as per-area video_maps

### LINKED ISSUE #812: In new facility config scheme, stars_maps should be revisited
A single value for it doesn't make sense for most facilities; e.g. N90 has it as
```
        "stars_maps": [
            "LGA-HPN ARR/DEP",
            "Coast",
            "NY Class B",
            "MVA",
            "LGA RNAV",
            "NE HELI RNAV ROUTES",
            "HPN 16/34 GPS FIXES",
            "NK78 GPS",
            "LGA LOC 31",
            "R5001/HELI GPS"
        ],
```
which is LGA-specific.

We should probably follow the real-world system and allow multiple (named) video map sets and then require that each local control position have one of them assigned.

mmp:
Video maps are now specified per-area in area_configs rather than as a facility-wide stars_maps field, allowing each TRACON area to have its own set of video maps. For ARTCC configs, we now have "eram_maps".

Map center/range can also be specified per-area in area_configs, with priority: controller_configs > area_configs > top-level config. Scenario default_maps take priority over area-level default_maps.

Fixes #812.


mmp:
Good feedback. Done and merged!

Issue #824, opened 2026-03-03
improve traffic call "looking": followups after some time if not seen, also double check if close but no cigar


mmp:
> if you get “looking” response, then I’d say if real pilots don’t see it within 30sec to 1min they’d most likely ask “where’s that traffic for AAL123?” Or “request update on that traffic for AAL123” or “we still don’t have the traffic, AAL123”.

> Also for “looking response”, if the traffic is not in the reported  “o’clock quadrant but “close” you might hear “is that the traffic at our 1 o’clock”? Or “we have traffic at out 1 o’clock insight”.

PR #864, opened 2026-04-16
Traffic Advisory Update if not seen. Wrong "O'Clock" quadrant added. 

Jud6969:
Referenced from issue #824 

Add two pilot-response variants:

- Wrong-quadrant: when no traffic matches the reported o'clock but there's real traffic within 5 NM / 1000 ft whose actual bearing is within 2 hours of the reported position, the pilot reports it at the corrected clock position ("is that the traffic at our 4 o'clock" / "traffic at our 4 o'clock in sight"). Sets TrafficInSight.
- Follow-up after 30–60 s of "looking" with no sighting: the pilot continously asks for an update ("where's that traffic" / "request update on that traffic" / "we still don't have the traffic"). Fires once per advisory.

### DIFF
--- a/aviation/intent.go
+++ b/aviation/intent.go
@@ -889,16 +889,18 @@ func (a ATISIntent) Render(rt *RadioTransmission, r *rand.Rand) {
 type TrafficAdvisoryResponse int
 
 const (
-	TrafficResponseIMC          TrafficAdvisoryResponse = iota // In IMC, can't see traffic
-	TrafficResponseLooking                                     // No traffic visible, will look
-	TrafficResponseTrafficSeen                                 // Traffic is in sight
-	TrafficResponseAcknowledged                                // Other traffic is maintaining visual separation
+	TrafficResponseIMC           TrafficAdvisoryResponse = iota // In IMC, can't see traffic
+	TrafficResponseLooking                                      // No traffic visible, will look
+	TrafficResponseTrafficSeen                                  // Traffic is in sight
+	TrafficResponseAcknowledged                                 // Other traffic is maintaining visual separation
+	TrafficResponseWrongQuadrant                                // Pilot sees traffic, but at a different o'clock than reported
 )
 
 // TrafficAdvisoryIntent represents a pilot's response to a traffic advisory
 type TrafficAdvisoryIntent struct {
 	Response               TrafficAdvisoryResponse
 	WillMaintainSeparation bool // If true, add "will maintain visual separation"
+	ActualOclock           int  // For TrafficResponseWrongQuadrant: the o'clock at which the pilot actually sees the traffic
 }
 
 func (t TrafficAdvisoryIntent) Render(rt *RadioTransmission, r *rand.Rand) {
@@ -915,6 +917,9 @@ func (t TrafficAdvisoryIntent) Render(rt *RadioTransmission, r *rand.Rand) {
 		}
 	case TrafficResponseAcknowledged:
 		rt.Add("[roger|copy the traffic|roger, we have the traffic]")
+	case TrafficResponseWrongQuadrant:
+		rt.Add("[traffic at our {num} o'clock in sight|we have traffic at our {num} o'clock]",
+			t.ActualOclock)
 	}
 }
 
--- a/sim/aircraft.go
+++ b/sim/aircraft.go
@@ -44,8 +44,9 @@ type SeenAircraft struct {
 }
 
 type UnseenTrafficCall struct {
-	Callsign   av.ADSBCallsign
-	CalledTime Time
+	Callsign         av.ADSBCallsign
+	CalledTime       Time
+	WhereAskFireTime Time // If non-zero and passed, pilot proactively asks "where's that traffic"
 }
 
 type Aircraft struct {
--- a/sim/commands.go
+++ b/sim/commands.go
@@ -490,39 +490,27 @@ func (s *Sim) handleTrafficAdvisory(ac *Aircraft, oclock int, miles int, traffic
 		return av.TrafficAdvisoryIntent{Response: av.TrafficResponseIMC}
 	}
 
-	// Convert o'clock to heading offset from aircraft heading
-	// 12 o'clock = 0 degrees, 3 o'clock = 90 degrees, etc.
-	oclockHeading := math.MagneticHeading((oclock % 12) * 30) // 0, 30, 60, 90... 330
-	trafficHeading := math.NormalizeHeading(ac.Heading() + oclockHeading)
+	kind, trafficFound, actualOclock := s.findAdvisoryTraffic(ac, oclock, miles, trafficAltFeet)
 
-	// Calculate the approximate position of the reported traffic
-	nmPerLong := ac.NmPerLongitude()
-	magVar := ac.MagneticVariation()
-	trafficPos := math.Offset2LL(ac.Position(), math.MagneticToTrue(trafficHeading, magVar), float32(miles), nmPerLong)
-
-	// Search for actual traffic near the reported position
-	// Tolerance: +/- 2 miles horizontal, +/- 1000 feet vertical
-	const horizontalToleranceNM = 2.0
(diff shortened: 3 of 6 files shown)


sim/control.go
```diff
@@ -2171,15 +2189,75 @@ func (s *Sim) handleTrafficAdvisory(ac *Aircraft, oclock int, miles int, traffic
...
 	ac.TrafficLookingUntil = s.State.SimTime.Add(time.Duration(10+s.Rand.Intn(10)) * time.Second)
+	ac.TrafficWhereAskAt = s.State.SimTime.Add(time.Duration(30+s.Rand.Intn(31)) * time.Second)
 	return av.TrafficAdvisoryIntent{Response: av.TrafficResponseLooking}
 }
 
+// findWrongQuadrantTraffic looks for real traffic within ~5 NM and ±1000 ft of
+// the advised altitude whose actual o'clock is within ±2 hours of the reported
+// one (but not matching it). Returns the callsign and the actual o'clock.
+func (s *Sim) findWrongQuadrantTraffic(ac *Aircraft, reportedOclock int, trafficAltFeet float32) (av.ADSBCallsign, int, bool) {
+	const maxRangeNM = 5.0
```
mmp: This feels high, like if a traffic call gives 1nm, pilots shouldn't report seeing traffic 6nm out. What about a sliding scale, like at minimum 1 mile and then like 25% of the issued distance, plus or minus?

sim/control.go
```diff
@@ -2171,15 +2189,75 @@ func (s *Sim) handleTrafficAdvisory(ac *Aircraft, oclock int, miles int, traffic
...
+	const oclockTolerance = 2
+
+	var bestCS av.ADSBCallsign
+	var bestOclock int
+	bestDist := float32(999999)
+
+	nmPerLong := ac.NmPerLongitude()
+	magVar := ac.MagneticVariation()
+
+	for cs, other := range s.Aircraft {
```
mmp: We already have a loop over aircraft where we compute distance, etc., in handleTrafficAdvisory. How about unifying into a single loop to reduce redundant work (and extra lines of code), probably in a utility function called by handleTrafficAdvisory). Then that function could either return in-sight traffic, or the closest candidate traffic (with a bool or somesuch indicating which it is.)

sim/control.go
```diff
@@ -2171,15 +2189,75 @@ func (s *Sim) handleTrafficAdvisory(ac *Aircraft, oclock int, miles int, traffic
...
+		}
+		if math.Abs(other.Altitude()-trafficAltFeet) > verticalToleranceFeet {
+			continue
+		}
+
+		// Actual o'clock of `other` relative to ac's nose.
+		bearingTrue := math.Heading2LL(ac.Position(), other.Position(), nmPerLong)
+		bearingMag := math.TrueToMagnetic(bearingTrue, magVar)
+		rel := math.NormalizeHeading(float32(bearingMag) - float32(ac.Heading()))
+		actualOclock := math.HeadingAsHour(rel)
```
mmp: You could use `math.HeadingDifference`, skip the conversion to o'clock, and then also not have to deal with the logic for `diff` below, then just check for headingDifference < 45 or what have you.

sim/control.go
```diff
@@ -2171,15 +2189,75 @@ func (s *Sim) handleTrafficAdvisory(ac *Aircraft, oclock int, miles int, traffic
...
+	nmPerLong := ac.NmPerLongitude()
+	magVar := ac.MagneticVariation()
+
+	for cs, other := range s.Aircraft {
+		if cs == ac.ADSBCallsign {
+			continue
+		}
+
+		dist := math.NMDistance2LL(ac.Position(), other.Position())
+		if dist > maxRangeNM {
```
mmp: Shouldn't the check be with respect to the distance issued in the traffic call, not an absolute distance? This will reject an aircraft 9nm away even if the traffic call was for 10 miles.

sim/control.go
```diff
@@ -2205,10 +2283,35 @@ func (s *Sim) checkDelayedTrafficInSight(ac *Aircraft) {
...
 }
 
+// checkTrafficWhereAsk fires once, 30–60 s after a "looking" reply, if the
+// pilot still hasn't spotted the traffic and no new advisory has superseded
+// it — the pilot proactively asks the controller for an update.
+func (s *Sim) checkTrafficWhereAsk(ac *Aircraft) {
+	if ac.TrafficWhereAskAt.IsZero() {
+		return
+	}
+	if ac.TrafficInSight {
```
mmp: I think this is buggy as written: if the aircraft previously reported some other traffic in sight, this will be true, even though it's not for the traffic that was called but not originally seen.

sim/control.go
```diff
@@ -2205,10 +2283,35 @@ func (s *Sim) checkDelayedTrafficInSight(ac *Aircraft) {
...
+		return
+	}
+
+	// Fire once, whether or not we actually have a frequency to talk on.
+	ac.TrafficWhereAskAt = Time{}
+	if ac.ControllerFrequency == "" {
+		return
+	}
+
+	s.enqueuePilotTransmission(ac.ADSBCallsign, TCP(ac.ControllerFrequency), PendingTransmissionTrafficWhere)
```
mmp: What happens if the aircraft has landed (for example)? Maybe at least condition this on the aircraft not having been deleted?

mmp:
(Also, please run `gofmt -w .` from the top-level for consistent code formatting.)

sim/control.go
Jud6969: Added a slider to the function. More miles in the advisory, the bigger the tolerance, and window which the aircraft will find the given traffic through.

	rangeTolNM := float32(reportedMiles) * 0.25
	if rangeTolNM < 1 {
		rangeTolNM = 1
	}
	minRangeNM := float32(reportedMiles) - rangeTolNM
	if minRangeNM < 0 {
		minRangeNM = 0
	}
	maxRangeNM := float32(reportedMiles) + rangeTolNM

sim/control.go
Jud6969: Dropped the o'clock conversion stuff, and switched to math.HeadingDifference against a 15 to 60 degree window. sim/control.go:2302-2303

sim/control.go
Jud6969: - Collapsed both passes into one helper findAdvisoryTraffic which does vertical filtering, the horizontal-proximity check, and the sliding-range + bearing check (for wrong-quadrant) inside a signl loop over s.Aircraft.(sim/control.go:2283-2309)
- Tagged the result rather than returning an ambiguous bool. Added a enum trafficMatchKind with three values: trafficMatchExact, trafficMatchWrongQuadrant, trafficMatchNone (sim/control.go 2233-2240). The helper returns (kind, callsign, actualOclock); switches on kind sim/control.go:2166.

sim/control.go
Jud6969: That should be fixed as now the search window increases depending on the miles given in the advisory.

sim/control.go
Jud6969: I removed those four lines. Other areas of the code already clear TrafficWhereAskAt on the same line that sets TrafficInSight.

sim/control.go
Jud6969: added continue sim/sim.go:1655 so deleted aircraft cant fall through checkTrafficWhereAsk(ac)
also added a line so it checks if the reported aircraft has a frequency. if they dont then it clears (sim/control.go 2361-2367)

Jud6969:
code should be formatted, from my end it is at least, let me know it if it is not

Jud6969:
Converting this to a draft since there were a lot of changes to the traffic advisory functions.

Jud6969:
updated with the main branch. did some testing by putting two aircraft next to each other going the same altitude and direction, then had one guy behind trying to find one of the aircraft. testing worked.

Jud6969:
It's been forever since I uploaded this and this is not 100%. I might come back to this later but for now I'm closing it.

Issue #781, opened 2026-01-29
(rare) aircraft with incorrect altitude from transponders


mmp:
Simulate (GA) aircraft with transponders that report incorrect altitudes -> need to tell them to stop squawking altitude. These should then report "reaching 3,000'" or what have you when they change altitudes -> use the STARS commands for pilot reported altitudes

PR #868, opened 2026-04-20
Add Realistic Altimeter Simulation

Jud6969:
Closes #852 , & #781

Okay this one is kind of huge so let me break it down:
- Pilots now have their own altimeter setting value, and indicated altitude (real altitude - altimeter bias)
- Pilots now have an altimeter bias. If the set altimeter is wrong, the pilot will indicate being at said altitude, but the radar will show the difference. (e.g. altimeter is 30.23 and they have 29.92 set at 10,000ft. The pilot will actually fly 310ft higher then their assigned altitude: 10,310ft displayed on the radar)
- `AltimeterSetting` is changed via the `ASXXXX` (altimeter command), issuing the ATIS, or both (e.g. Orlando Information K current, altimeter 29.92). **(NOT TESTED YET)**
- FL180 is the transition level. Aircraft climbing past or to FL180 will automatically switch to 29.92 on their altimeter. Likewise if they are descending below FL180. Aircraft above FL180 and descending will queue an assigned altimeter (arrival airport altimeter by default if they have the ATIS) **(NOT TESTED YET)**

To keep it simple (for me), heres the logic:
- Indicated altitude = what the pilot has tuned with their altimeter. Altitude - altimeterBias (local - set * 1000)
- Mode C Altitude = true altitude - the pressure difference (standard - pressure) * 1000. This in the code is only used when the aircraft is above FL180 which brings the idea of having a lowest usable flight level. If the pressure is below 29.92 inHg FL180 is not usable. meaning the lowest usable flight level is FL190 or above depending on how low the pressure is. This follows real world logic.

How this is configured for aircraft
- VFR aircraft will always have the correct altimeter.
- (assuming local did their job) departures will automatically have the correct altimeter
- Arrivals will always have the correct altimeter. I might make this to where if they did not call with the ATIS there is a chance that they have an incorrect altimeter setting.

Push coming soon. I am testing this on my own machine. I will reopen this as a draft soon.


Jud6969:
I'm going to bring this back up to date and rework this with better testing and structure.

mmp:
Great, sounds good!

PR #871, opened 2026-04-22
Pilots requesting to maintain altitude, when one is not assigned.

### LINKED ISSUE #775: ask for altitude if vectored off CVS/DVS and not given an altitude
"what altitude do you want us at?", etc.

jessie846:
Closes #775


mmp:
Nice, merged!

PR #872, opened 2026-04-23
Shared Datablock Manipulations Across Same TCWs / Added Shared Scope Feature

### LINKED ISSUE #736: should see same TCW (data block positions, P cones, J rings, etc.) when sharing a scope
[discord](https://discord.com/channels/1090685318014976081/1453536844732956924/1453536844732956924)

Jud6969:
Closes #736. I added a bit more to it.
Per-ACID track annotation sync (always on)
                                                                                                                                                                                                                                                                                                     
When two or more humans are at the same TCW, annotations placed on a specific aircraft propagate to every controller at that TCW:

  - J-ring radius, ATPA cone length
  - Leader-line direction (and FDAM variant), UseGlobalLeaderLine
  - Force-full-datablock, per-ACID PTL
  - TPA-size, ATPA monitor, ATPA warn/alert toggles
  - Display-requested-altitude, LDB beacon-code
  - Stale entries auto-pruned when a track leaves.

 Opt-in scope-prefs sync ("Sync Scope Setup")

A checkbox on the Join-as-Relief dialog. When any relief ticks it, the TCW-wide flag flips on (sticky for the session) and scope preferences round-trip between the primary and every participating relief via a single opaque JSON blob:

  - Scope center, range, range-ring radius, rotation
  - Brightness levels, PTL length
  - Video map selections, weather levels
  - SSA filters + GI text, list positions
  - LDR direction and every other configurable DCB feature

Per-controller opt-out: a relief who leaves the checkbox off at join time does not participate in sync, even if another relief already enabled it TCW-wide — their scope stays local-only.

Excluded (remain strictly per-user): CharSize, AudioVolume, AudioEffectEnabled, DwellMode, AutoCursorHome, CursorHome, DisplayDCB, DCBPosition, RestrictionAreaSettings.

 **Architecture**

- Server owns TCWDisplayState per TCW, lazily created on first signon. Not persisted across sim restarts.
- Delivery: RPC mutations bump a revision counter; every state-update poll carries the current blob. State-update poll cap tightened from 1 s → 100 ms so observers update within a frame.
- Push throttle on the client side caps outgoing RPCs at 10 Hz so an active pan/drag doesn't flood the server.
- Rejoin inherits the TCW's current state; last-leaves-then-rejoin preserves annotations for the session.


sim/tcw_display.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// ScopePrefsBlob is an opaque JSON-encoded STARS Preferences
+	// snapshot: the whole struct sans a handful of per-user fields
+	// (character size, audio, cursor home, dwell mode, DCB position).
+	// The server does not inspect the blob -- it is written by the
+	// client that last mutated a scope-wide setting and echoed back
+	// to all controllers at the TCW so they can apply it to their
+	// local prefs. Zero-length means no client has seeded shared
+	// state yet.
+	ScopePrefsBlob []byte
```
mmp, line 55: I thought the idea was that these things would not be synced to other controllers. What's the point of this? (Same for ScopePrefsRev)

stars/track_shared.go
mmp, line 1: I don't think this warrants a new file; let's add to stars/track.go instead.

stars/track_shared.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/mmp/vice/panes"
+	"github.com/mmp/vice/sim"
+)
+
+// SharedTrackAnnotationFields is the authoritative list of
+// sim.TrackAnnotations field names that are synced across relief
+// controllers. Keep in sync with the struct in sim/tcw_display.go; a
+// reflective test guards against drift.
+var SharedTrackAnnotationFields = []string{
```
mmp: I'm not sure I understand why we need to enumerate these as strings here (or have the methods below)

stars/ui.go
```diff
@@ -530,9 +530,10 @@ func (sp *STARSPane) DrawInfo(c *client.ControlClient, p platform.Platform, lg *
 		candidateHolds := make(map[string]av.Hold)
 		ps := sp.currentPrefs()
 		ctr := util.Select(ps.UseUserCenter, ps.UserCenter, ps.DefaultCenter)
+		rng := ps.Range
```
mmp: why this change?

stars/track.go
```diff
@@ -37,26 +37,14 @@ type TrackState struct {
...
 
 	IsSelected bool // middle click
 
-	// We handed it off, the other controller accepted it, we haven't yet
-	// slewed to make it a PDB.
-	DisplayFDB bool
-
 	// Hold for release aircraft released and deleted from the coordination
 	// list by the controller.
 	ReleaseDeleted bool
```
mmp: Shouldn't this also be included so that it matches on a second scope?

More generally, I feel like just about everything in TrackState should be under the umbrella of what's managed with this change.

(And further, that the non-exported fields: track, trackTime, etc., should perhaps be exported and managed server side.)

stars/cmdtools.go
```diff
@@ -1090,95 +1098,128 @@ func registerToolsCommands() {
...
+			ctx.Client.SetTrackJRingRadius(acid, 0, cb) // can't have both
+		}
 		return nil
 	})
 
 	// 6.21.8 Delete TPA Cone for single track (Implied command) (p. 6-178)
-	registerCommand(CommandModeNone, "*P[SLEW]", func(state *TrackState) {
-		state.ConeLength = 0
+	registerCommand(CommandModeNone, "*P[SLEW]", func(sp *STARSPane, ctx *panes.Context, trk *sim.Track) {
+		if trk.IsAssociated() {
```
mmp: I believe that it should be possible to put a p-cone on an unassociated VFR track.

server/manager.go
```diff
@@ -282,6 +287,12 @@ type JoinSimRequest struct {
 	Password        string
 	Privileged      bool
 	JoiningAsRelief bool
+	// SyncScopeState is a client-side hint (opt-in on the Join-as-Relief
+	// dialog) — when true, the client mirrors the per-TCW ScopeView
+	// (range/pan/range-ring) rather than its local preference. Server
+	// doesn't use it; it's round-tripped so the caller's own client has
+	// the intent and passes it to NewControlClient.
+	SyncScopeState bool
```
mmp: I'm sort of inlined to make this automatic and not opt-in/opt-out: if we're simulating a new controller sitting down at the same radar scope, then that's how it works there.

server/dispatcher.go
```diff
@@ -1102,3 +1102,210 @@ func (sd *dispatcher) AnnotateFlightStrip(args *AnnotateFlightStripArgs, _ *stru
...
+	Value           *bool
+}
+
+type SetTrackLeaderLineArgs struct {
+	ControllerToken string
+	ACID            sim.ACID
+	Direction       *math.CardinalOrdinalDirection
+}
+
+const SetTrackJRingRadiusRPC = "Sim.SetTrackJRingRadius"
```
mmp: These are a lot of RPCs to add; I'm wondering if there's a simpler approach:

- Make the current stars.TrackState be managed server-side in the sim package. Maybe it's even folded into the sim.Track structure?
- Client code that wants to change it (add a p-cone, whatever), uses a single new RPC e.g. "Sim.UpdateTrackState" that takes a callsign and the updated TrackState.
- There is a small risk of a race condition if both the regular user and the relief controller update something in the track state at the same time, then only one of those changes may stick. This is probably fine.

client/client.go
```diff
@@ -303,7 +319,7 @@ func (c *ControlClient) GetUpdates(eventStream *sim.EventStream, p platform.Plat
 	}
 
 	// Wait in seconds between update fetches; no less than 50ms
-	rate := math.Clamp(1/c.State.SimRate, 0.05, 1)
+	rate := math.Clamp(0.1/c.State.SimRate, 0.05, 0.1)
```
mmp: I think this will 10x the amount of network bandwidth used, all the time?!?

Also, per the CL description I think the idea was low-latency updates when one user drags the scope. I think it's fine for that to be updated once per second.

sim/tcw_display.go
Jud6969: ScopePrefsBlob is there for storing any changes between two controllers on a shared TCW if a relief controller opted into "syncing their scope" with the other person. So when they first load in, settings are matched, then whatever changes while the controllers occupy the same TCW, that data gets stored here, then the ScopePrefsRev gets bumped by 1 to indicate that there is newer data, clients look at the data, then sync the local prefs.

This wasn't of the original issue. I thought it would be cool for simulating "looking at the same TCW" IRL. If the coding is too tacky then I can remove it.

stars/track_shared.go
Jud6969: This is dead. This was for testing. Forgot to remove it

stars/ui.go
Jud6969: Sorry, this was caught in the crossfire of me having to revert a change I made. Reverted.

stars/track.go
Jud6969: DisplayFDB is synced, it was moved. Reference sim.TrackAnnotations.

I will work on having the rest of the TrackState stuff managed under the server.  TrackHistory will still be Client-side since that would be bandwith intensive.

server/dispatcher.go
Jud6969: New commit moved all the RPCs to one. Kept the possibly race sensitive RPCs in place: all the MSAWs, and SetTrackInQLRegion. So now there's only 7 compared to what it would've been.

server/manager.go
Jud6969: I have switched it to automatic, let me know what you think of it when you test. This now forces the scope sync instead of letting the user choose.

client/client.go
Jud6969: sorry lol. switched back.

stars/cmdtools.go
Jud6969: fixed. switched track associated to trk.ADSBCallsign

stars/track.go
Jud6969: switched

Jud6969:
New commit should've resolved all your comments. Let me know if there's anything else.

Jud6969:
I'm switching this back to a draft I'm not confident that this is complete or fully functioning. I have to update a lot since there have been a lot of commits since I've last tried this with someone else.

Jud6969:
This was a lot. I will come back to this later. I'm thinking I could bring this in smaller chunks:
- Shared datablock manipulations for relief controllers (how it is in CRC currently)
- Shared Pref Sets & Scope movement (something controllers can opt into)
This also isn't 100%. I will close this and maybe come back to this later.

PR #878, opened 2026-04-29
Fix Aircraft Randomly Leaving The Hold

### LINKED ISSUE #876: aircraft leaving hold without being told to
[discord](https://discord.com/channels/1090685318014976081/1497439181310394650/1497439181310394650)

> I noticed a bug (atleast to the best of my knowledge) when controlling Newark when you give the command to hold at PENNS, the aircraft does a full loop of the published hold, but then goes off in a direction of maybe around 040. not sure why, but it wasn't just one plane, it was all of them.  I have not tested other holding patterns. Here is an attached video of the timelapse where I am issuing holds, and they don't properly stay holding if that helps:

Jud6969:
Closes #876 

HoldStateTurningInbound bailed out of the inbound turn at 60 degrees and fell back to ShouldTurnToIntercept which is written for localizer intercepts. When the turn was more than 10 degrees it returned MajorOvershoot, which falls back to "fly present heading" which leads to the behaviour of the aircraft randomly leaving the hold as stated in the issue Now it just completes the 180 degree turn through the inbound and and has FlyingInbound turn direct-to-fix as it already does. 

The tradeoff with this fix is that the aircraft will overshoot just a bit on the inbound turn. It shouldn't be that big of a deal though as its barely noticeable. 

Tested on BOS FV 4L/R with the same conditions stated in the bug report and it works flawlessly now.

### DIFF
--- a/nav/procedures.go
+++ b/nav/procedures.go
@@ -538,18 +538,10 @@ func init() {
 		HoldStateTurningInbound: func(nav *Nav, fh *FlyHold, wxs wx.Sample, simTime Time) (math.MagneticHeading, av.TurnDirection, HoldState) {
 			turn := util.Select(fh.Hold.TurnDirection == av.TurnRight, av.TurnRight, av.TurnLeft)
 
-			if math.HeadingDifference(nav.FlightState.Heading, fh.Hold.InboundCourse) > 60 {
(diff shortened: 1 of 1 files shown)


mmp:
nice!

PR #895, opened 2026-05-11
nav/wx: use arrival airport METAR for approach speed wind additive

radarContacto:
Use METAR wind data (including gusts) from the arrival airport when computing approach speed, instead of the weather sample at the aircraft's current position. Caches decoded METARs per airport in the wx.Model to avoid repeated fetches.


mmp:
Looks good!

PR #896, opened 2026-05-11
sim: add optional scenario description displayed in new sim window

radarContacto:
Adds a "description" field to scenario JSON that scenario developers can use to describe goals, tips, or context for the scenario. When present, it is shown in the new simulation selection window below the landing runways section in its own auto-sized bordered box.


mmp:
Looks good, thanks!

PR #902, opened 2026-06-03
Update sayfix.json

GooseLived:
Expanded updates of pronunciation recognition for numerous I90 TRACON fixes that were mispronounced or not recognized.  Included SID fixes and IAP fixes for IAH and HOU.


mmp:
Thanks!

PR #904, opened 2026-06-11
Updated Y90 scenario & configuration

CiceroIsBack:
- changed scope chars, callsigns, and frequencies to their correct values
- updated departure headings to account for the airport flow
- added additional realistic departure exit points
- added simple "Sequencing Practice" scenario for BDL
- removed old/bad route

There's a lot more work to do on this facility, but this is my first attempt at facility engineering, so I figured I'd keep this first try small. 
Please let me know if there's something I should've done differently--I want to learn!

### DIFF
--- a/resources/configurations/ZBW/Y90.json
+++ b/resources/configurations/ZBW/Y90.json
@@ -16,12 +16,12 @@
       "radio_name": "Bradley Approach"
     },
     "1Z": {
-      "callsign": "BDL_W_APP",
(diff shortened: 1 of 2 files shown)


mmp:
Nice! This looks reasonable to me. CC @radarContacto who I believe has some familiarity with Y90 and may have some feedback.

mmp:
Thanks!

PR #905, opened 2026-06-14
Fix crash on Apple M5 by bumping go-m1cpu to v0.2.1

jordw:
## Problem

vice crashes on launch on Apple **M5** machines. The cause is the indirect dependency `github.com/shoenig/go-m1cpu`, pinned at `v0.1.6`, which segfaults / hits a `CFRelease` on a NULL reference on Apple Silicon it doesn't recognize.

Dependency chain:

```
util -> github.com/shirou/gopsutil/v3/cpu -> github.com/shoenig/go-m1cpu
```

## Fix

Bump `go-m1cpu` to **v0.2.1**:

- **v0.2.0** — adds M4 support and fixes the segmentation fault on M5 processors
- **v0.2.1** — fixes the `CFRelease NULL crash on M5 Pro/Max`

Only `go.mod` / `go.sum` change. `go mod tidy` is clean and the affected packages build.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -87,7 +87,7 @@ require (
 	github.com/power-devops/perfstat v0.0.0-20210106213030-5aafc221ea8c // indirect
 	github.com/randall77/makefat v0.0.0-20210315173500-7ddd0e42c844 // indirect
 	github.com/rivo/uniseg v0.4.7 // indirect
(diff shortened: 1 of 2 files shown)


mmp:
Thanks!

PR #906, opened 2026-06-14
Keep the OS cursor visible over imgui UI

jordw:
## Problem

The OS mouse cursor disappears over imgui UI where it shouldn't:

- the main **menu bar**, and
- the **floating imgui windows** (Messages, Flight Strips).

The cursor correctly hides only over the STARS radar scope, where the pane draws its own cursor — that part is by design.

## Cause

A pane (STARS over the radar scope) requests `imgui.MouseCursorNone` so it can draw its own cursor. imgui resets its cursor to `Arrow` every frame (`ImGui::NewFrame`), so `MouseCursorNone` reflects a pane actively drawing its own cursor *this* frame. But the OS-cursor handling in `platform/glfw.go` hid the cursor for that state unconditionally:

```go
} else if g.mouseDeltaMode || imgui_cursor == imgui.MouseCursorNone {
    g.window.SetInputMode(glfw.CursorMode, glfw.CursorHidden)
}
```

This was latent until #cb9323da ("ui: move Messages and Flight Strips to floating imgui windows") introduced imgui windows overlaying the scope — before that the whole main window was the scope.

## Fix

Only hide the OS cursor when imgui isn't using the mouse:

```go
} else if !io.WantCaptureMouse() && (g.mouseDeltaMode || imgui_cursor == imgui.MouseCursorNone) {
    g.window.SetInputMode(glfw.CursorMode, glfw.CursorHidden)
}
```

Otherwise show the standard cursor, falling back to the arrow when imgui has no shape to show — which also avoids indexing `mouseCursors` with the out-of-range `MouseCursorNone` (-1).

This is independent of window focus, so a pane's own cursor still renders over its window when the application is unfocused.

Result:
- Over the scope (`!WantCaptureMouse`) → still hidden, pane draws its own cursor. ✔
- Over the menu bar / floating windows → OS cursor visible. ✔
- Focused or unfocused → consistent. ✔

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/platform/glfw.go
+++ b/platform/glfw.go
@@ -348,6 +348,7 @@ func (g *glfwPlatform) NewFrame() {
 
 	// Mouse cursor
 	imgui_cursor := imgui.CurrentMouseCursor()
+	io := imgui.CurrentIO()
 
 	if g.cursorOverride != nil {
 		// A pane set a specific OS cursor (e.g., ERAM); show it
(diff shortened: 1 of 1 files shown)


jordw:
For context, I just got an M5 Macbook Pro. This didn't seem happen on Sequoia, but does happen on Tahoe.

mmp:
Amusingly(?) I (and claude) reworked the STARS cursor code this morning to make it a little less hacky and prevent it from hiding the cursor when the mouse was over the window but vice wasn't the active application. (This was with an M4 and Sequoia, though.)

Would you mind syncing TOT and see how things are for you with that fix, a37f551e1b1c4a1388c00c31b217ab9dfefc6922? Happy to merge this as well if it's still needed.

jordw:
Will do!

jordw:
Confirmed that this is no longer needed, thanks!

PR #924, opened 2026-07-03
ERAM NEXRAD

checkandmate1:
Finally!

Did a few things to make this work, and I updated some of the backend precip stuff:

- Switched the precip source to IEM NEXRAD n0q.
Looks like this IEM source is behind the [NOAA NEXRAD Map](https://www.ncei.noaa.gov/maps/radar/), which seems to be very similar to what CRC displays.

- Added ERAM NEXRAD


radar/weather.go
```diff
@@ -91,75 +139,81 @@ func (w *WeatherRadar) HaveWeather() [NumWxLevels]bool {
...
+			return 0
+		}
+		return s.LevelForDBZ(precip.DBZ[idx])
+	}
 
-		// Process each row of the precipitation grid
+	out := make([]*renderer.CommandBuffer, s.NumLevels())
+	for level := 1; level <= s.NumLevels(); level++ {
+		tb.Reset()
+		any := false
```
mmp: Since `any` is a type in go even though this is legal it's a little distracting. `haveWx` or similar instead?

wx/precip.go
```diff
@@ -50,31 +50,101 @@ func (p Precip) BoundsLL() math.Extent2D {
...
+// square side length in nautical miles; resolution is fixed at 2 px/NM.
+func MakeDebugPrecip(center math.Point2LL, sideNM int) *Precip {
+	const pixelsPerNM = 2
+	resolution := sideNM * pixelsPerNM
+	dbz := make([]byte, resolution*resolution)
+
+	half := float32(resolution) / 2
+	radius2 := half * half
+	thirdY := resolution / 3
+	for y := 0; y < resolution; y++ {
```
mmp: `for y := range resolution` svp

wx/precip.go
```diff
@@ -50,31 +50,101 @@ func (p Precip) BoundsLL() math.Extent2D {
...
+		switch {
+		case y < thirdY:
+			v = 55 // severe → Extreme
+		case y < 2*thirdY:
+			v = 45 // medium → Heavy
+		default:
+			v = 35 // low → Moderate
+		}
+		dy := float32(y) - half + 0.5
+		for x := 0; x < resolution; x++ {
```
mmp: `for x := range resolution`

wx/precip.go
```diff
@@ -161,57 +226,53 @@ func estimateDBZ(root *kdNode, rgb [3]byte) float32 {
...
 
 	// Determine the dBZ for each pixel.
 	ny, nx := img.Bounds().Dy(), img.Bounds().Dx()
 	dbzImage := make([]byte, nx*ny)
 	for y := range ny {
 		for x := range nx {
 			px := rgba.RGBAAt(x, y)
-			dbz := estimateDBZ(root, [3]byte{px.R, px.G, px.B})
+			dbz := float32(-100)
+			if !noData(px) {
```
mmp: `!noData` -> "not no data" is a little convoluted with the double negative. How about changing `noData` to `haveData` (and flipping the logic in the implementations) and then `if haveData(px)` here?

eram/prefs.go
```diff
@@ -200,6 +205,17 @@ const (
 	Line4Type
 )
 
+// NexradLevel toolbar button settings. Non-OFF values match the digits the
+// NX LVL button displays (e.g., 23 = Heavy and Extreme). OFF is intentionally
+// non-zero so the JSON zero value (from preference sets that predate this
+// field) is reserved as "uninitialized" and gets migrated to the default.
+const (
+	NexradLevelOff     = -1
```
mmp: These are a bit confusingly named given `nexradLevelModerate` et al. above-this had me confused about some of the code in `nexradLevelLabel` at first. How about including "toolbar" in these names and/or naming like `nexradDisplayLevel*` in the other ones to better distinguish them?

mmp (commented):
Yaay! Overall looks good. Here are some picky things from a first quick pass through. Given the holiday weekend it may be a day or two before I can make a more thorough pass through.

eram/toolbar.go
mmp, line 2400: Do we need to handle WX toolbar tearoffs here (not sure)?

eram/nexrad.go
```diff
@@ -0,0 +1,138 @@
...
+	return fmt.Sprintf("%d", level)
+}
+
+func handleNexradLevelClick(ep *ERAMPane, pref *int) {
+	mouse := toolbarDrawState.mouse
+	if mouse == nil {
+		return
+	}
+
+	idx := 0
```
mmp: could write all this (I think) as `idx := slices.Index(nexradLevelCycle, *pref)`

eram/nexrad.go
```diff
@@ -0,0 +1,138 @@
...
+var nexradLevelCycle = []int{NexradToolbarAll, NexradToolbarHeavy, NexradToolbarExtreme, NexradToolbarOff}
+
+func nexradLevelLabel(level int) string {
+	if level == NexradToolbarOff {
+		return "OFF"
+	}
+	return fmt.Sprintf("%d", level)
+}
+
+func handleNexradLevelClick(ep *ERAMPane, pref *int) {
```
mmp, line 120: In CRC, it looks like left click once it's OFF leaves it at OFF and shows the circled E cursor briefly if you click again. (And then middle click stops at 123 and shows the circled P if you middle click again. Here we seem to wrap around in those cases. (I'm guessing CRC is right since that matches how we handle other toolbar up/down value adjustments.)

eram/nexrad.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+	}
+	return p
+}()
+
+// ERAM NEXRAD colors. Brightness is applied at draw time via
+// ps.Brightness.NEXRAD.
+var (
+	nexradModerateColor = renderer.RGB{R: 0, G: 0, B: 1}
+	nexradHeavyColor    = renderer.RGB{R: 0, G: 1, B: 1}
```
mmp, line 46: Just confirming, same color for heavy and extreme is intentional?

eram/toolbar.go
```diff
@@ -497,6 +497,52 @@ func (ep *ERAMPane) drawToolbarMenu(ctx *panes.Context, scale float32) {
 			toolbarDrawState.occlusionExtent = oc
 		}
 
+	case toolbarWX:
+		if toolbarDrawState.lightToolbar != [4][2]float32{} {
```
mmp, line 501: Minor, but when the WX toolbar is open, CRC looks like:

Image

while vice looks like:

Image

So, missing the "ATC Tools" on the left and, a blue background vs salmon background for the WX toolbar, and then an extra column on the right in vice.

mmp (commented):
Looks great, a few last minor things.

I have realized we have an existing bug with not accounting for the fact that the STARS wx fetches are at 0.5nm resolution but ERAM is 1nm--I think this is why you had to make those ARTCC radius values so large. I can fix that up after this lands.

eram/nexrad.go
checkandmate1: Yeah. Heavy is the same color as extreme - it's just checkered.

mmp:
Looks good! I will restart the scraper to pick this up.

PR #933, opened 2026-07-21
Add support for real-world scheduled traffic

planevan:
This pull request adds support for generating traffic from real-world airline schedules.

Instead of relying entirely on randomly generated departures and arrivals, Vice can now load scheduled departures and arrivals from CSV files and generate traffic that closely follows real-world airline operations.

The initial implementation includes a built-in MSP weekday schedule as a proof-of-concept implementation in the M98 Scenario. 

Features

- Adds a Real World Schedules traffic source.
- Loads schedules from CSV files.
- Supports built-in schedule catalogs.
- Matches scheduled flights to available airlines and aircraft types.
- Preserves existing random traffic generation for airports without schedules.
- Adds realistic departure gate/taxi timing and queuing for scheduled departures.
- Preserves existing random traffic generation when real-world schedules are not used.

Testing

Tested using the included MSP weekday schedule.

Verified:

- Schedule loading
- Arrival generation
- Departure generation
- Realistic departure timing
- Multiple traffic sessions without crashes

The feature compiles successfully, existing tests pass, and simulator testing showed significantly more realistic traffic flow compared to immediate scheduled departures.

This is intended as a foundation for additional airports and schedules. Eventual additions will include the ability for users to create their own schedules by simply creating an airport file in the schedules folder and dropping their CSV in there. Currently, a schedule.json needs to be used. 

Future improvements may include automatic schedule discovery, additional built-in schedules, further refinement of departure sequencing, and arrivals spawning inside TRACON to simulate taking over from another controller.


sim/schedule_catalog_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+}
+
+func TestBundledMSPDevelopmentSchedule(t *testing.T) {
+	catalog, err := LoadBuiltInScheduleCatalog(util.GetResourcesFS(), "schedules")
+	if err != nil {
+		t.Fatalf("LoadBuiltInScheduleCatalog: %v", err)
+	}
+
+	schedule, ok := catalog.Find("KMSP", "development_test")
```
mmp: Running tests with `go test ./...` is currently failing--I think either `development_test` here needs to be `summer_weekday`, or we need another schedule checked in.

resources/schedules/KMSP/schedule.json
```diff
@@ -0,0 +1,13 @@
+{
+  "airport": "KMSP",
```
mmp: `"airport": "KMSP"` is somewhat redundant since the airport name is in the folder name. What about a) renaming this file to `schedules.json` and then b) having the contents be like:
```json
[
    {
      "id": "summer_weekday",
      "name": "MSP Summer Weekday",
      "file": "summer_weekday.csv",
      "description": "Full-day real-world MSP schedule.",
      "timezone": "America/Chicago"
    }
  ]
```

resources/schedules/README.md
mmp, line 1: Rather than documenting it here, could you add a short section to `website/facility-engineering.html`? Maybe a new section after "Arrivals and Overflights" and before"VFR Aircraft" under "Scenario Groups"?

It would also be nice to mention there that routes are taken from existing routes in the scenario file if the departure/destination airports match up.

sim/traffic_provider.go
```diff
@@ -0,0 +1,313 @@
...
+// departure times are treated as pushback times; a random 10-20 minute taxi-out
+// duration is generated once per departure when the provider is first used.
+// The schedule clock starts at ScheduleStartMinute when the provider is created.
+type scheduleTrafficProvider struct {
+	airport string
+	start   Time
+
+	departures            []scheduledDeparture
+	nextDeparture         int
+	departuresInitialized bool
```
mmp: I believe this is unused and could be deleted.

sim/spawn.go
```diff
@@ -370,7 +409,6 @@ func (s *Sim) addDepartureToPool(ac *Aircraft, runway av.RunwayID, manualLaunch
 	depState := s.DepartureState[ac.FlightPlan.DepartureAirport][runway]
 	if ac.FlightPlan.Rules == av.FlightRulesIFR {
 		if manualLaunch {
-			// Keep them moving and for HFR, request the release immediately.
```
mmp, line 373: Please restore this comment.

sim/spawn_arrivals.go
```diff
@@ -141,7 +165,6 @@ func (s *Sim) createArrivalNoLock(group string, arrivalAirport string) (*Aircraf
 	nasFp.SecondaryScratchpad = arr.SecondaryScratchpad
 	nasFp.RNAV = s.State.FacilityAdaptation.Datablocks.DisplayRNAVSymbol && arr.IsRNAV
 
-	// For ERAM, set AssignedAltitude and derive PerceivedAssigned from waypoint restrictions.
```
mmp, line 144: This and a few following comments got lost, please restore.

sim/spawn_departures.go
```diff
@@ -713,11 +821,10 @@ func (s *Sim) createIFRDepartureNoLock(departureAirport string, runway av.Runway
 		nasFp.EntryFix = ac.FlightPlan.DepartureAirport
 	}
 	nasFp.ExitFix = shortExit
-	if dep.Scratchpad != "" { // this has top priority
+	if dep.Scratchpad != "" {
 		nasFp.Scratchpad = dep.Scratchpad
 	} else if sp1 := s.State.FacilityAdaptation.Datablocks.Scratchpad1; sp1.DisplayExitFix ||
 		sp1.DisplayExitFix1 || sp1.DisplayExitGate || sp1.DisplayAltExitGate {
-		// Don't set the scratchpad; it will be set automatically.
```
mmp, line 720: Another lost comment.

sim/spawn_departures.go
```diff
@@ -728,30 +835,24 @@ func (s *Sim) createIFRDepartureNoLock(departureAirport string, runway av.Runway
...
 
-	ac.HoldForRelease = (ap.HoldForRelease || exitRoute.HoldForRelease) && ac.FlightPlan.Rules == av.FlightRulesIFR // VFRs aren't held
+	ac.HoldForRelease = (ap.HoldForRelease || exitRoute.HoldForRelease) && ac.FlightPlan.Rules == av.FlightRulesIFR
 	s.assignDepartureController(ac, &nasFp, ap, exitRoute, departureAirport, string(runway))
 
 	if err := s.assignSquawk(ac, &nasFp); err != nil {
 		return nil, err
 	}
 
-	// Departures aren't immediately associated, but the STARSComputer will
```
mmp, line 738: Lost comment here and then two more right below.

sim/spawn_departures.go
```diff
@@ -728,30 +835,24 @@ func (s *Sim) createIFRDepartureNoLock(departureAirport string, runway av.Runway
 	nasFp.AssignedAltitude = util.Select(!isTRACON, ac.FlightPlan.Altitude, 0)
 	nasFp.RNAV = s.State.FacilityAdaptation.Datablocks.DisplayRNAVSymbol && exitRoute.IsRNAV
 
-	ac.HoldForRelease = (ap.HoldForRelease || exitRoute.HoldForRelease) && ac.FlightPlan.Rules == av.FlightRulesIFR // VFRs aren't held
```
mmp, line 731: Lost comment

sim/traffic_provider.go
```diff
@@ -0,0 +1,313 @@
...
+	"errors"
+	"fmt"
+	"sort"
+	"time"
+
+	av "github.com/mmp/vice/aviation"
+	"github.com/mmp/vice/util"
+)
+
+// trafficProvider supplies automatically generated IFR aircraft to the
```
mmp, line 21: This is really nice, introducing an interface here for this.

sim/traffic_provider.go
```diff
@@ -0,0 +1,313 @@
...
+// simulation. It also controls when the next departure request should occur;
+// random traffic uses a rate-based delay while schedule traffic uses the next
+// published pushback time plus simulated taxi-out time.
+type trafficProvider interface {
+	createIFRDeparture(s *Sim, airport string, runway av.RunwayID) (*Aircraft, time.Duration, error)
```
mmp, line 26: For consistency, let's either include "IFR" in both of these method names or in neither of them. (Either way's fine with me.)

sim/traffic_provider.go
```diff
@@ -0,0 +1,313 @@
...
+		return false
+	}
+
+	if percentage == 100 || flight.Cargo {
+		return true
+	}
+
+	// Use a stable hash so the same percentage consistently selects the same
+	// flights each time the scenario is loaded.
+	hash := uint32(2166136261)
```
mmp, line 83: We have a `HashString64()` function in `util`; could just replace this with e.g. `return int(HashString64(key)%100) < percentage`

cmd/vice/simconfig.go
mmp, line 2539: Currently the local time in the UI is shown as "00:00" (unless it's manually edited). I think that if we have `if TrafficSoruce == RealWorldSchedule` and then do something like
```
  loc, err := time.LoadLocation(schedule.Timezone)
  local := c.StartTime.In(loc)
  lc.ScheduleStartMinute = local.Hour()*60 + local.Minute()
  c.scheduleStartTimeText = formatScheduleStartTime(lc.ScheduleStartMinute)
```
it will stay in sync.

sim/traffic_provider.go
```diff
@@ -0,0 +1,313 @@
...
+
+	sort.SliceStable(p.arrivals, func(i, j int) bool {
+		return p.arrivals[i].offset < p.arrivals[j].offset
+	})
+
+	return p
+
+}
+
+// initializeDepartures generates one taxi-out duration for each scheduled
```
mmp, line 151: stale comment?

sim/schedule_catalog.go
```diff
@@ -0,0 +1,221 @@
...
+	"fmt"
+	"io/fs"
+	"path"
+	"sort"
+	"strings"
+)
+
+const scheduleCatalogFilename = "schedule.json"
+
+// BuiltInSchedule describes one schedule distributed in Vice's resources.
```
mmp, line 15: How about renaming this to `FlightSchedule` (and similarly for `BuiltInScheduleCatalog` and `BuiltInScheduleSummary`)? The rationale being that "built-in" is sort of implied and "flight" more accurately describes what it's holding.

server/scenario.go
```diff
@@ -2473,6 +2483,13 @@ func LoadScenarioGroups(extraScenarioFilename string, extraVideoMapFilename stri
...
 
+	scheduleCatalog, err := sim.LoadBuiltInScheduleCatalog(util.GetResourcesFS(), "schedules")
+	if err != nil {
+		e.Error(err)
+	} else {
+		attachBuiltInSchedules(catalogs, scheduleCatalog)
+	}
+
 	lg.Infof("LoadScenarioGroups total: %s", time.Since(start))
 	return scenarioGroups, catalogs, mapSpecs, briefs, extraScenarioErrors
```
mmp, line 2494: Let's also return `scheduleCatalog` here, so that we don't end up re-loading the schedules in `newScheduleTrafficProvider`. In that way, we load and validate schedules just once, which matches how the scenario loading flow works currently for other things. (Generally, we do all the error checking for scenarios at initial loading time and then the rest of the code can assume that things are reasonable, since we error out and exit if there are scenario errors.)

Roughly, if we then store the full schedule catalog in `SimManager`, then pass the selected schedule (if any) to `NewSim` in `sim.NewSimConfiguration`, then we can store it directly in the `Sim`. Then `Sim.activeTrafficProvider()` can be simplified to just pass the schedule to `newScheduleTrafficProvider` without needing to load it.

sim/schedule.go
```diff
@@ -0,0 +1,248 @@
...
+	case originMatches && !destinationMatches:
+		return ScheduleOperationDeparture
+	case destinationMatches && !originMatches:
+		return ScheduleOperationArrival
+	default:
+		return ScheduleOperationUnknown
+	}
+}
+
+func normalizeScheduledAircraftType(value string) string {
```
mmp, line 61: Just thinking out loud, should these rewrites be encoded in a JSON file in `resources` so that people doing facility engineering can add them? Should we just require that the schedule CSV files have these rewrites already done? This partially connects to the question of where the data comes from and how we want to be importing it.

mmp (commented):
This is a great addition and I think will be a huge step forward for the realism of vice. Lots of comments below, some small things, some bigger changes, but overall I think the general form of this is quite good.

sim/schedule_catalog_test.go
planevan: this looks like a stale reference from when the bundled schedule was renamed. I’ll check the intent of the test, but I believe updating development_test to summer_weekday is the correct fix rather than adding a duplicate schedule.

resources/schedules/KMSP/schedule.json
planevan: Good point. I'll refactor it to use a schedules.json catalog within each airport directory.

sim/schedule_catalog_test.go
planevan: I just realized that I fixed this yesterday when I cleaned up that development stuff. So this should be fixed already.

resources/schedules/README.md
planevan: I moved the schedule documentation into facility-engineering.html, added a new "Real World Schedules" section (including the route reuse behavior), added it to the navigation sidebar, and reduced the README to a pointer to the engineering documentation.

sim/traffic_provider.go
planevan: oops This was left over from an earlier version of the traffic provider and wasn't being used anymore. Removed.

sim/spawn.go
planevan: Restored!

sim/spawn_arrivals.go
planevan: Restored these. My bad.

sim/spawn_departures.go
planevan: Restored!

sim/spawn_departures.go
planevan: Restored.

sim/spawn_departures.go
planevan: Restored.

planevan (commented):
Restored comments.

sim/traffic_provider.go
planevan: Thank you!

PR #939, opened 2026-08-03
Adjust ERAM Datablock Leader Line Alignment

Jud6969:
I noticed in vice, ERAM data blocks were misaligned with the leader line, causing it to look off, and sometimes clipping with elements of the data block. Below are examples.
image 
image 
image

Basically, instead of the geometry needing to be tuned by per-direction pixel constants (which drifted depending on font size), now the placement is derviced from the font metrics, and the data block contents.

I made all my tests in reference to CRC ERAM. They all look pretty accurate now. Here are some comparison screenshots:

East Owned

**Before**
image
**After** 
image
**CRC**
Screenshot 2026-08-03 020803

Northeast Owned

**Before**
image
**After**
image
**CRC**
image

East Not Owned

**Before**
image
**After**
image
**CRC**
image

Specifics:
  - the line meets the CID row's ink top for N/E/W/NE/NW, and the callsign row's ink bottom for S/SE/SW
  - 6 pixels of clearance between the line and the block, in every direction, produced by offsetting the block rather than shortening the line
  - the VCI and "R" symbols sit a third of a character clear of the main columns, matching CRC's 6-pixel gap rather than the normal 2
  - the "R" shortens the line only when the leader line runs into it (when the data block is set to northeast, and east).
  - vertical leaders meet the block under the "R" rather than beside it
  - leader lines are nudged half a pixel onto a pixel boundary so a 2-wide line covers two rows cleanly instead of smearing across three

Also fixes a mismatch where non-owned FDB leader lines were scaled to 0.7 in drawLeaderLines but not in datablockAnchor, leaving those blocks positioned for a full-length line. 

I tested this on a windows machine, on a 4k monitor. Please let me know what thoughts you have on these changes.

### DIFF
--- a/eram/datablock.go
+++ b/eram/datablock.go
@@ -97,29 +97,36 @@ func dbChopTrailing(f []dbChar) []dbChar {
 func dbDrawLines(lines []dbLine, td *renderer.TextDrawBuilder, pt [2]float32,
 	font *renderer.Font, sb *strings.Builder, brightness radar.Brightness,
 	dir math.CardinalOrdinalDirection, halfSeconds int64) {
-	scale := float32(dbLineOffsetScale) // 1.5 for default font
-	if len(lines) >= 5 {
-		if lines[3].ch[0].ch != rune('R') {
-			scale = dbLineOffsetScale
-		}
-	}
 	glyph := font.LookupGlyph(' ')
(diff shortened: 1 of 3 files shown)


mmp:
Nice!

PR #942, opened 2026-08-04
Update airport-timezones.json

GooseLived:
Added addition EST airports for IND TRACON. Removed duplicated from "Indianapolis" Time zone as this doesn't actually exist.


mmp:
Thanks!

PR #965, opened 2026-08-23
Add ZTL Area 1 low sector scenarios

gagostin1:
Adds standalone scenarios for the ZTL Area 1 low sectors:

Moped 47 (L47)
Shine 44 (L44)
Bristol 45 (L45)

Includes sector-specific arrival/departure flows, handoff logic, altitude procedures, and traffic based on the current ZTL SOP/LOAs and real-world routings.

All three scenarios were tested locally in Vice for traffic flow, handoffs, altitude behavior, and handoff deletion.


mmp:
Nice!

PR #966, opened 2026-08-25
p80: update to current charts, fix runway assignment and go-arounds

nreisbeck:
- fix KRATRWest_Final expect_approach: pilots were assigned the ILS 28L while the scratchpad showed I8R; the east flow and the north-side downwind geometry both point to the north runway
- TMBRS4: west transition FLOWR..WIDMR..WHEAT..KOLCH replaces the old ROAGE leg; east adds HAZEE and the PUBBB flyover, replacing MYCRO
- KRATR3: BOOSR/RALEE legs west, VEEGN/NNUTT east, replacing LIQWD/SSSUN/SHYNE and VOODU
- HHOOD6: charted 230K at BEEVS
- add spoken STAR labels to the KRATR, TMBRS, and OCITY flows so pilots announce the arrival on check-in
- drop revision numbers from flow keys (HHOOD, KRATR, TMBRS, OCITY), following the existing HELNS key; the spoken labels carry the current revision so future chart updates no longer require key renames
- add go_around_assignments: the final-only configurations (KP1/KP2) had neither go-around nor departure assignments, so the go-around controller lookup fell through to the current controller (the tower) and go-arounds never returned to the final controller. KP1/KP2 now send go-arounds to 1F; the combined configurations get the per-runway split from the P80 SOP (north runway to NORTH, south to HOOD)

### DIFF
--- a/resources/configurations/ZSE/P80.json
+++ b/resources/configurations/ZSE/P80.json
@@ -44,13 +44,19 @@
   "facility_adaptations": {
     "configurations": {
       "KPD": {
+        "go_around_assignments": {
+          "KPDX/28R": "1N",
+          "KPDX/10L": "1N",
+          "KPDX/28L": "1D",
+          "KPDX/10R": "1D"
+        },
         "inbound_assignments": {
           "HELNS": "1N",
-          "HHOOD5East": "1N",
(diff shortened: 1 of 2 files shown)


mmp:
Nice!

PR #967, opened 2026-08-26
visual approaches: direct the numbers, clearances from the downwind

### LINKED ISSUE #880: Add support for "proceed|turn direct the numbers" (before or after visual approach clearance)
“Turn (left/right (optional)*) direct the numbers, contact tower”
“Turn direct the numbers, cleared visual approach…”
“Cleared visual approach runway 4L, turn direct to the numbers”

### LINKED ISSUE #882: Allow clearing aircraft on a downwind for a visual approach
Currently we more or less depend on intersecting the visual approach route. In such cases we should probably just set them up so they fly direct to a suitable fix that is behind them and in from there.

nreisbeck:
Two visual-approach improvements:

- **Direct the numbers / direct the field** (#880): `DTN` (and `LDTN`/`RDTN` with a turn direction) sends the aircraft direct to the assigned approach runway's threshold; `DTF` direct to the arrival airport. Spoken forms: "turn left direct the numbers", "proceed direct the field". Works before or after a visual approach clearance.

- **Visual clearances from the downwind** (#882): when the abeam projection onto the reference route is behind the aircraft, the clearance now joins the route at a point downwind of the aircraft — it continues ahead, turns in to that fix, and flies the route inbound from there — rather than refusing. This follows the approach suggested in the issue.

Also rewords the synthesis-failure refusal from "we don't know runway X" (which reads as a data problem) to "we're not in position for the visual to runway X".

Fixes #880
Fixes #882

### DIFF
--- a/nav/commands.go
+++ b/nav/commands.go
@@ -1310,6 +1310,33 @@ func (nav *Nav) DistanceAlongRoute(fix string) (float32, error) {
 	return distance, nil
 }
 
+// DirectRunwayNumbers has the pilot proceed visually direct to the
+// assigned approach runway's threshold ("direct the numbers").
+func (nav *Nav) DirectRunwayNumbers(turn av.TurnDirection, simTime Time, delayReduction time.Duration) av.CommandIntent {
+	ap := nav.Approach.Assigned
+	if ap == nil {
+		return av.MakeUnableIntent("unable. you haven't given us a runway")
+	}
+	wps := []av.Waypoint{
+		{Fix: "_NUMBERS", Location: ap.Threshold},
+		nav.FlightState.ArrivalAirport,
+	}
+	nav.EnqueueDirectFix(wps, turn, simTime, delayReduction)
+	nav.Approach.NoPT = false
+	nav.Approach.InterceptState = NotIntercepting
+	return av.DirectNumbersIntent{Turn: turn}
+}
+
+// DirectAirport has the pilot proceed direct to the arrival airport
+// ("proceed direct the field").
+func (nav *Nav) DirectAirport(turn av.TurnDirection, simTime Time, delayReduction time.Duration) av.CommandIntent {
+	wps := []av.Waypoint{nav.FlightState.ArrivalAirport}
+	nav.EnqueueDirectFix(wps, turn, simTime, delayReduction)
+	nav.Approach.NoPT = false
+	nav.Approach.InterceptState = NotIntercepting
+	return av.DirectAirportIntent{Turn: turn}
+}
+
 func (nav *Nav) ResumeOwnNavigation() av.CommandIntent {
 	if nav.Heading.Assigned == nil {
 		// This is a weird response but keeping the original behavior
--- a/sim/command_parser.go
+++ b/sim/command_parser.go
@@ -608,6 +608,13 @@ func (s *Sim) runOneControlCommand(tcw TCW, callsign av.ADSBCallsign, command st
 		}
 
 	case 'D':
+		if command == "DTN" {
+			// Direct the (runway) numbers.
+			return s.DirectRunwayNumbers(tcw, callsign, av.TurnClosest, delayReduction)
+		} else if command == "DTF" {
+			// Direct the field/airport.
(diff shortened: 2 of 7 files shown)


nreisbeck:
hold for bug fix

stt/handlers.go
```diff
@@ -861,6 +861,29 @@ func registerAllCommands() {
...
+		func() string { return "LDTN" },
+		WithName("left_direct_numbers"), WithPriority(14),
+	)
+	registerSTTCommand(
+		"turn|proceed right direct [to] the numbers",
+		func() string { return "RDTN" },
+		WithName("right_direct_numbers"), WithPriority(14),
+	)
+	registerSTTCommand(
+		"turn|proceed [direct] direct [to] the numbers",
```
mmp, line 877: Here and the next one has `[direct] direct` which seems redundant and above just has `direct`; I assume they should all be the same? IMHO `[direct]` is probably best as far as being a bit more forgiving, which is helpful with STT since the transcripts are often a little garbled.

stt/handlers.go
```diff
@@ -861,6 +861,29 @@ func registerAllCommands() {
 	)
 
 	// === NAVIGATION COMMANDS ===
+	// Direct the numbers / the field (must outrank the generic {fix}
+	// patterns so "the numbers" isn't fed to the fix matcher).
+	registerSTTCommand(
```
mmp, line 866: Please add entries in the "ATC Commands (Spoken)" section in `website/index.html` to document these.

sim/command_parser.go
```diff
@@ -736,7 +743,11 @@ func (s *Sim) runOneControlCommand(tcw TCW, callsign av.ADSBCallsign, command st
 		}
 
 	case 'L':
-		if len(command) >= 5 && command[1] == 'D' {
+		if command == "LDTN" {
```
mmp, line 746: Please add entries in the "ATC Commands (Keyboard)" section in `website/index.html` to document these. Also please add them to `secondaryAcCommands` in `cmd/vice/ui.go` so they're there in the app's help window.

nav/commands.go
```diff
@@ -1310,6 +1310,33 @@ func (nav *Nav) DistanceAlongRoute(fix string) (float32, error) {
...
 
+// DirectRunwayNumbers has the pilot proceed visually direct to the
+// assigned approach runway's threshold ("direct the numbers").
+func (nav *Nav) DirectRunwayNumbers(turn av.TurnDirection, simTime Time, delayReduction time.Duration) av.CommandIntent {
+	ap := nav.Approach.Assigned
+	if ap == nil {
+		return av.MakeUnableIntent("unable. you haven't given us a runway")
+	}
+	wps := []av.Waypoint{
+		{Fix: "_NUMBERS", Location: ap.Threshold},
```
mmp, line 1321: I think the Waypoint here should have `Flags: av.WaypointFlagLand` (so that the aircraft is deleted when it reaches this point or goes around if it's still goo high) as well as an altitude restriction based on the runway elevation and threshold crossing height, roughly
```
rwy := ...
alt := rwy.Elevation + rwy.ThresholdCrossingHeight
wp.SetAltitudeRestriction(MakeAtAltitudeRestriction(float32(alt)))
```

nav/commands.go
```diff
@@ -1310,6 +1310,33 @@ func (nav *Nav) DistanceAlongRoute(fix string) (float32, error) {
...
+	nav.EnqueueDirectFix(wps, turn, simTime, delayReduction)
+	nav.Approach.NoPT = false
+	nav.Approach.InterceptState = NotIntercepting
+	return av.DirectNumbersIntent{Turn: turn}
+}
+
+// DirectAirport has the pilot proceed direct to the arrival airport
+// ("proceed direct the field").
+func (nav *Nav) DirectAirport(turn av.TurnDirection, simTime Time, delayReduction time.Duration) av.CommandIntent {
+	wps := []av.Waypoint{nav.FlightState.ArrivalAirport}
```
mmp, line 1333: (Same comment as above w.r.t. flags and altitude restriction.)

mmp (commented):
Sorry for not getting to this sooner. Looks generally good! A few comments below.

